PR #63, opened 2026-07-08
consensus: count Dilithium sigops for P2MR (witness v2)

augchan42:
WitnessSigOps() returned 0 for witness versions other than v0, so P2MR Dilithium signature operations were never counted toward MAX_BLOCK_SIGOPS_COST. A block full of P2MR inputs could be disproportionately expensive to validate.

This parses the P2MR leaf script and counts OP_CHECKSIGDILITHIUM* / OP_CHECKMULTISIGDILITHIUM* using the existing DILITHIUM_SIGOP_COST constant from script.h (50).

One open question for maintainers: 50 caps a block at 1600 Dilithium sigops (80000/50), which binds before the 8 MB block size does (~2100 single-sig P2MR inputs). If the intended multiplier is closer to the ~10x verification-cost ratio vs ECDSA, that's a one-line change to script.h — but it should be a deliberate decision since it sets Dilithium throughput per block. Follows up on #24.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -2403,28 +2403,17 @@ size_t static WitnessSigOps(int witversion, const std::vector<unsigned char>& wi
 
     if (witversion == 2 && witprogram.size() == WITNESS_V2_P2MR_SIZE && witness.stack.size() >= 2) {
         // P2MR witness: [args...] [script] [control_block] [optional annex]
-        // Determine if annex is present (last element starts with ANNEX_TAG)
-        size_t last = witness.stack.size() - 1;
-        bool has_annex = (witness.stack.size() >= 2 &&
-                          !witness.stack[last].empty() &&
-                          witness.stack[last][0] == ANNEX_TAG);
-        size_t control_idx = has_annex ? last - 1 : last;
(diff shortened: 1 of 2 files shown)


augchan42:
Reviewed 7c843d2af — LGTM, this is better than the original approach. Verified GetSigOpCount(true) applies DILITHIUM_SIGOP_COST per op and keys×cost for multisig with accurate OP_N counts, so delegation loses nothing; the annex/control indexing mirrors VerifyWitnessProgram; and accurate key counts replace the flat MAX_PUBKEYS_PER_MULTISIG overcount, matching P2WSH precedent.

Built the branch head and ran sigopcount_tests: the new cases all pass individually, but the full-suite run shows 3 failures. Root cause is pre-existing, not this PR: GetTxSigOpCost (sigopcount_tests.cpp:197, from 57dd34052) hard-asserts a WITNESS_PUBKEYTYPE expectation that no longer holds — master fails the same way (2 failures). Because it's a C assert, the abort skips fixture teardown and the next cases die on duplicate ArgsManager::AddArg registrations, which is why they pass standalone. d99c74b8c on fix/p2mr-merge-master-audit already aligns that expectation — worth cherry-picking into this branch (or merging separately) so the suite goes green here.

PR #65, opened 2026-07-09
p2p: reduce MAX_PROTOCOL_MESSAGE_LENGTH from 70 MB to 9 MB

augchan42:
The inherited 70 MB cap let a single peer force allocation of very large message buffers. 9 MB still clears the largest legitimate message (8 MB serialized block, MAX_BLOCK_SERIALIZED_SIZE) with headroom.

Safe for mixed-version P2P on the running testnet: no legitimate message exceeds 8 MB today, so old and new nodes interoperate through the upgrade.

### DIFF
--- a/src/net.h
+++ b/src/net.h
@@ -63,8 +63,8 @@ static constexpr std::chrono::minutes TIMEOUT_INTERVAL{20};
 static constexpr auto FEELER_INTERVAL = 2min;
 /** Run the extra block-relay-only connection loop once every 5 minutes. **/
 static constexpr auto EXTRA_BLOCK_RELAY_ONLY_PEER_INTERVAL = 5min;
-/** Maximum length of incoming protocol messages (raised to carry ≥64 MB blocks). */
-static const unsigned int MAX_PROTOCOL_MESSAGE_LENGTH = 70 * 1000 * 1000; // 70 MB
+/** Maximum length of incoming protocol messages (block size + overhead). */
+static const unsigned int MAX_PROTOCOL_MESSAGE_LENGTH = 9 * 1000 * 1000; // 9 MB
 /** Maximum length of the user agent string in `version` message */
 static const unsigned int MAX_SUBVERSION_LENGTH = 256;
 /** Maximum number of automatic outgoing nodes over which we'll relay everything (blocks, tx, addrs, etc) */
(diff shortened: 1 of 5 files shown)


augchan42:
Provenance for the 70 MB value: Bitcoin's original cap was 4 MB. ea17bb60a (Aug 2025) raised it to 70 MB alongside the 64 MiB block cap — the rationale was max block + ~6 MiB slack for framing. When 2b44ec5b7 (Feb 2026) reduced blocks to 8 MB, the message cap was left behind. This PR applies the same max-block-plus-slack formula (8 MB + 1 MB) to the current block size, so it's really the missing half of the February block-size reduction rather than a new policy.

Oskii:
## Audit review

### What was checked
- `MAX_BLOCK_SERIALIZED_SIZE` is 8_000_000; weight math implies any valid with-witness block serializes to ≤ 8 MB, so a 9 MB transport cap (8 MB + 1 MB) is above the largest legitimate `block` / near-full `blocktxn`.
- Cap is enforced in `V1Transport` via `hdr.nMessageSize > MAX_PROTOCOL_MESSAGE_LENGTH` (`src/net.cpp`), and stays ≤ `MAX_SIZE` (32 MiB) so that check is not a no-op.
- Deriving the cap from `MAX_BLOCK_SERIALIZED_SIZE` (already on the branch) is correct; hardcoding 9_000_000 would drift if the block size changes again.

### Defect found and fixed
The branch redefined `MAX_PROTOCOL_MESSAGE_LENGTH` as `MAX_BLOCK_SERIALIZED_SIZE + 1*1000*1000`, which **broke** `test/lint/lint-btq-consensus-constants.py`:

```
KeyError: 'MAX_BLOCK_SERIALIZED_SIZE'
```

Pushed commit `fd25ee839`:
- teach the lint evaluator named identifiers and assert `8_000_000 + 1_000_000`
- add `net_tests/max_protocol_message_length` constant lock
- clarify the net.h comment (payload-only; inv headroom)

### Test evidence
| Check | Result |
|-------|--------|
| `python3 test/lint/lint-btq-consensus-constants.py` on tip before fix | **FAIL** (`KeyError`) |
| same after fix | **PASS** |
| `test_btq --run_test=net_tests/max_protocol_message_length` | **PASS** |
| `test_btq --run_test=net_tests/v2transport_test` | **PASS** |
| `test/functional/p2p_invalid_messages.py` (incl. at-limit accept / oversize disconnect) | **PASS** |

Safe for mixed-version testnet: no legitimate message exceeds 8 MB today, so old peers sending ≤8 MB blocks still interoperate; oversized junk is rejected earlier.

PR #66, opened 2026-07-09
p2p: add NODE_BTQ_DILITHIUM service bit

augchan42:
Advertises bit 26 in local services so BTQ nodes are identifiable in addr gossip and crawler/seed tooling can distinguish them from other Bitcoin-lineage nodes.

Scope note: this is identification only — GetDesirableServiceFlags does not require the bit from peers, and cross-chain handshakes are already blocked by BTQ's message-start bytes. Requiring the bit for peer selection could be a follow-up once most of the network advertises it.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -787,7 +787,7 @@ namespace { // Variables internal to initialization process only
 int nMaxConnections;
 int nUserMaxConnections;
 int nFD;
-ServiceFlags nLocalServices = ServiceFlags(NODE_NETWORK_LIMITED | NODE_WITNESS);
+ServiceFlags nLocalServices = ServiceFlags(NODE_NETWORK_LIMITED | NODE_WITNESS | NODE_BTQ_DILITHIUM);
 int64_t peer_connect_timeout;
 std::set<BlockFilterType> g_enabled_filter_types;
 
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -200,6 +200,7 @@ static std::string serviceFlagToStr(size_t bit)
     case NODE_COMPACT_FILTERS: return "COMPACT_FILTERS";
     case NODE_NETWORK_LIMITED: return "NETWORK_LIMITED";
     case NODE_P2P_V2:          return "P2P_V2";
+    case NODE_BTQ_DILITHIUM:   return "BTQ_DILITHIUM";
     // Not using default, so we get warned when a case is missing
     }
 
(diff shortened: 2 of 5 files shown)


augchan42:
Context on why this is worth a bit now rather than at fork time: our message-start bytes separate BTQ from Bitcoin, but they don't separate old-BTQ from new-BTQ — every version of our node shares the same magic. The pending tier-2 changes (LWMA activation, P2MR-only Dilithium, sigop counting) are hard forks, so pre- and post-fork nodes will peer freely while following different rules. The standard mechanism (cf. NODE_BITCOIN_CASH) is to advertise a service bit well ahead of the fork, then flip GetDesirableServiceFlags to prefer it at fork time — which only works if most of the network already advertises it by then. Retroactive advertisement doesn't work.

Secondary: DNS seeders can filter on it, and getpeerinfo showing BTQ_DILITHIUM makes stale infra nodes easy to spot.

If we decide not to gate peering at the fork, this is fair to close — advertising alone changes nothing.

Oskii:
## Audit review

### What was checked
- Flag is OR'd into `nLocalServices` in `src/init.cpp` only; `GetDesirableServiceFlags()` still returns `NODE_NETWORK|NODE_WITNESS` (or limited variant) and does **not** require `NODE_BTQ_DILITHIUM` — so mixed-version testnet peers without the bit still connect.
- `serviceFlagToStr` covers the new enum (RPC `localservicesnames`).
- Cross-chain isolation remains via message-start bytes; this bit is identification/crawler aid only, as claimed.

### Defect found and fixed
Original PR used **bit 26** (`1 << 26`). That bit is the experimental full-RBF / `NODE_REPLACE_BY_FEE` (`NODE_FULL_RBF`) allocation in Bitcoin Core lineage tooling. Colliding makes crawler/tooling semantics ambiguous.

Pushed commit moves the flag to **bit 25**, documents why, adds `netbase_tests/node_btq_dilithium_service_bit`, and mirrors the constant in `test_framework/messages.py`.

### Test evidence
| Check | Result |
|-------|--------|
| `test_btq --run_test=netbase_tests/node_btq_dilithium_service_bit` | **PASS** (value `1<<25`, string `BTQ_DILITHIUM`, desirable flags do not require it) |
| regtest `getnetworkinfo` smoke | `localservicesnames` includes `BTQ_DILITHIUM`; hex services `...02000409` (bit 25 set) |

Note: bit still sits in the 24–31 experimental range; a permanent BIP-allocated bit can be a follow-up if this becomes a peer-selection requirement.

PR #67, opened 2026-07-16
security: cleanse Dilithium seed temporaries in ext-key paths

augchan42:
Decode(), Derive(), and SetSeed() copy the 32-byte Dilithium seed into a heap std::vector before expanding it, and the vector was destroyed without zeroing — leaving seed material in freed memory for anyone with memory read access (core dumps, /proc/pid/mem). Cleanses each temporary after last use, matching the file's existing pattern.

Reworked from the original BTQ-AUDIT-082 fix: the v0.4.0 HD redesign already covered the HMAC buffers and packed sk, so only these three temporaries remained. Passes dilithium_key/wallet/basic suites locally.

### DIFF
--- a/src/crypto/dilithium_key.cpp
+++ b/src/crypto/dilithium_key.cpp
@@ -289,7 +289,9 @@ void CDilithiumExtKey::Decode(const unsigned char code[DILITHIUM_EXTKEY_SIZE])
 
     // Regenerate the expanded keypair from the seed.
     std::vector<unsigned char> seed_vec(seed.begin(), seed.end());
-    if (!key.GenerateFromEntropy(seed_vec)) {
+    const bool ok = key.GenerateFromEntropy(seed_vec);
+    memory_cleanse(seed_vec.data(), seed_vec.size());
+    if (!ok) {
         key = CDilithiumKey();
     }
 }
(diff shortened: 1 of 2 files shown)


Oskii:
## Audit review

### What was checked
- `Decode` / `Derive` / `SetSeed` each heap-copy the 32-byte seed into a `std::vector` for `GenerateFromEntropy`. Without cleansing, that seed survives in freed heap pages (core dumps, cold boot, etc.).
- HMAC `I[]` buffers on those paths were already cleansed (v0.4 HD redesign); the PR correctly targets the remaining `seed_vec` temporaries and cleanses them on both success and failure paths (`memory_cleanse` before the `if (!ok)` branch).
- Early-return path in `Decode` (corrupt master/child metadata) wipes `seed` in-place and never builds `seed_vec` — OK.

### Defect found and fixed
Same pattern existed in `DescriptorScriptPubKeyMan::GetNewDestination` (`scriptpubkeyman.cpp`): HMAC output → heap `seed` → `GenerateFromEntropy` → destroy without zeroing (only `I[]` was cleansed).

Pushed `0a89c95b7` to cleanse that descriptor seed temporary the same way.

### Test evidence
| Check | Result |
|-------|--------|
| `test_btq --run_test=dilithium_wallet_tests` | **PASS** |
| `test_btq --run_test=dilithium_key_tests` | **PASS** |
| `test_btq --run_test=dilithium_basic_tests` | **PASS** |

No other production `GenerateFromEntropy` call sites take a heap seed temporary without a post-use cleanse.

PR #68, opened 2026-07-16
security: constant-time comparison for Dilithium key material

augchan42:
memcmp and std::array::operator== short-circuit on the first differing byte, leaking through timing how many leading bytes of two secrets match. Adds a TimingSafeEqual helper (volatile accumulator, XOR) for CDilithiumKey/PubKey equality, ext-key fingerprints, and the challenge-hash check in the reference implementation's crypto_sign_verify_internal().

One site is new relative to the original BTQ-AUDIT-016/-173 finding: the v0.4.0 ext-key operator== compares the 32-byte master seed via std::array's short-circuiting compare — that's the most sensitive equality in the file and is now constant-time. Ordering comparisons (operator<) are unchanged; they're not equality oracles. Passes dilithium_key/wallet/basic suites locally.

### DIFF
--- a/src/crypto/dilithium/avx2/sign.c
+++ b/src/crypto/dilithium/avx2/sign.c
@@ -451,9 +451,15 @@ int crypto_sign_verify_internal(const uint8_t *sig, size_t siglen, const uint8_t
   shake256_absorb(&state, buf.coeffs, K*POLYW1_PACKEDBYTES);
   shake256_finalize(&state);
   shake256_squeeze(buf.coeffs, CTILDEBYTES, &state);
-  for(i = 0; i < CTILDEBYTES; ++i)
-    if(buf.coeffs[i] != sig[i])
+  {
+    /* Constant-time challenge comparison: no early exit on first mismatch.
+     * Not linked today (Makefile builds ref/), but keep avx2 in sync. */
+    volatile uint8_t result = 0;
+    for(i = 0; i < CTILDEBYTES; ++i)
+      result |= buf.coeffs[i] ^ sig[i];
+    if(result)
       return -1;
+  }
 
   return 0;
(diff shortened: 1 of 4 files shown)


Oskii:
## Audit review

### What was checked
- Secret equality sites: `CDilithiumKey` / `CDilithiumPubKey` / ext-key seed & fingerprint, plus the challenge-hash check in `crypto_sign_verify_internal`.
- `TimingSafeEqual` uses a volatile XOR accumulator (no early exit). Functional correctness can be unit-tested; absolute constant-time against optimizing compilers is not proven empirically here.

### Defects found and fixed
The original PR still left gaps:
1. **`CDilithiumExtKey::operator==`** used `&&` so a metadata mismatch could skip the seed/chaincode scan; `chaincode` still used `uint256`/`ChainCode` `operator==` (short-circuiting `memcmp`).
2. **`CDilithiumExtPubKey::operator==`** same chaincode issue.
3. Challenge compare in **ref** `sign.c` lacked `volatile` on the accumulator; **avx2** `sign.c` still had the early-exit loop (not linked today, but drift risk).
4. **No tests** locking functional equality behavior.

Pushed `3cb74cca3`: always CT-scan seed/chaincode/fingerprint into locals before combining, volatile accumulators in ref+avx2, and `dilithium_key_tests` coverage (including last-byte mismatch).

### Test evidence
| Check | Result |
|-------|--------|
| `test_btq --run_test=dilithium_key_tests` | **PASS** |
| `test_btq --run_test=dilithium_wallet_tests` | **PASS** |
| `test_btq --run_test=dilithium_basic_tests` | **PASS** |

Ordering operators (`operator<`) intentionally unchanged — not equality oracles for secrets.

PR #69, opened 2026-07-17
consensus: activate LWMA from block 1 on mainnet

augchan42:
Mainnet had a 300000-block legacy-retarget window before LWMA activation — manipulable exactly when the chain is weakest (low hashrate at launch). Mainnet is pre-launch, so activating from block 1 is free (BTQ-AUDIT-103).

Mainnet only, following the nDilithiumP2MRHeight height-gate pattern: testnet/signet/regtest keep the scheduled height, so no live-chain impact and no reset. The activation test now locks in both values; the two legacy-bounds pow_tests construct their legacy window explicitly since mainnet no longer has one. pow_lwma_tests and pow_tests pass locally.

Should land before the external audit snapshot — the auditors need to review the difficulty design as it will ship.

### DIFF
--- a/src/kernel/chainparams.cpp
+++ b/src/kernel/chainparams.cpp
@@ -90,7 +90,12 @@ public:
         consensus.powLimit = uint256S("00000377ae000000000000000000000000000000000000000000000000000000");
         consensus.nPowTargetTimespan = 14 * 24 * 60 * 60; // two weeks (legacy, pre-LWMA)
         consensus.nPowTargetSpacing = 1 * 60;
-        consensus.nLWMAHeight = 300000;
+        // BTQ-AUDIT-103: LWMA from block 1 so mainnet never has a pre-LWMA
+        // legacy-retarget window that can be manipulated. Mainnet only: the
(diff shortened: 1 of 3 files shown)


Oskii:
## Audit review

**Verdict: approve (no code changes).**

### What was checked
- Diff only changes `CMainParams` `nLWMAHeight` from `300000` → `1`. Testnet/signet/regtest values are untouched (`testnet` remains `300000`, locked by the updated unit test).
- Call sites of `nLWMAHeight` gate legacy retarget vs LWMA; activating at 1 removes the mainnet pre-LWMA window (BTQ-AUDIT-103 claim).
- Mainnet is pre-launch / not carrying a live history that would be rewritten by this height change (unlike testnet, which correctly keeps 300000).
- Legacy ±4× bounds tests now construct a local `nLWMAHeight = 300000` window so they still exercise the pre-LWMA path after mainnet’s activation moved to 1.

### Test evidence
| Check | Result |
|-------|--------|
| `pow_lwma_tests` (incl. mainnet=1 / testnet=300000) | **PASS** |
| `pow_tests` | **PASS** |

No amendments required.

Issue #60, opened 2026-06-09
test: inherited Bitcoin test vectors fail under BTQ params (bloom_tests, bip324_tests) and abort the unit suite


BarneyChambers:
## Summary

Two unit-test suites inherited from Bitcoin Core fail on `master` because they hardcode **Bitcoin** test vectors/keys that are invalid under BTQ's chain parameters. Because the first failure **aborts mid-test** (via `assert`), it skips the test fixture's `gArgs.ClearArgs()`, which then makes every subsequent fixture trip a duplicate-`AddArg` assertion — cascading into hundreds of spurious failures in a full `src/test/test_btq` run.

These are **pre-existing on `master`** (the test files are byte-identical to master; not introduced by any recent PR). Filing for triage so they can be fixed in a dedicated PR.

## Failure 1 — `bloom_tests/bloom_create_insert_key`

`src/test/bloom_tests.cpp` decodes a hardcoded Bitcoin **mainnet** WIF:

```cpp
std::string strSecret = std::string("5Kg1gnAjaLfKiwhhPpGS3QfRg2m6awQvaj98JCZBZQ5SuS2F15C");
CKey key = DecodeSecret(strSecret);
CPubKey pubkey = key.GetPubKey();   // aborts: keydata is null
```

Under BTQ's base58 `SECRET_KEY` prefix the WIF does not decode, so `key` is invalid and `CKey::GetPubKey()` hits `assert(keydata)` (`key.cpp:189`) → `SIGABRT`.

**Suggested fix:** make the key chain-parameter-independent — set it from the raw 32-byte secret (the bytes that WIF encodes) instead of `DecodeSecret`, preserving the same pubkey so the expected bloom-filter bytes remain valid. (Several other bloom sub-tests use the same hardcoded WIF.)

## Failure 2 — `bip324_tests/packet_test_vectors` (~280 assertions)

`src/test/bip324_tests.cpp` checks hardcoded BIP324 v2-transport vectors (session IDs, garbage terminators, ciphertext). They mismatch BTQ's transport/network parameters, e.g.:

```
check Span{out_session_id} == cipher.GetSessionID() has failed
check out_ciphertext == ciphertext has failed
```

**Suggested fix:** regenerate the vectors against BTQ's transport parameters, or gate/skip the suite if BTQ does not intend to maintain BIP324 v2 test-vector parity. (~280 vectors — non-trivial; needs the reference generator with BTQ params.)

## Cascade / harness impact

[...]

PR #82, opened 2026-07-26
rpc: report an error for legacy Dilithium addresses instead of tripping an assert

### LINKED ISSUE #81: rpc: validateaddress returns "Internal bug detected" for any legacy Dilithium address
On v0.4.2, `validateaddress` fails with an internal-bug error for every legacy (pre-BIP360) Dilithium address:

```
$ btq-cli validateaddress nEw8yjgSEg5hqRB6bgkbuQaajkzLc76BN4
error code: -1
error message:
Internal bug detected: isValid == error_msg.empty()
../../src/rpc/output_script.cpp:64 (operator())
```

Expected: `{"isvalid": false, "error": "..."}`.

Reproduces on regtest with any of the four legacy Dilithium address forms (base58 pubkey-hash and script-hash, bech32 witness-v0 key-hash and script-hash).

## Root cause

`validateaddress` asserts `isValid == error_msg.empty()` (`src/rpc/output_script.cpp:64`) — an invalid destination must always come with an explanatory error.

29a82ac1a ("wallet+consensus: make Dilithium P2MR-only end-to-end") changed `ValidDestinationVisitor` so legacy Dilithium destinations return `false` (`src/addresstype.cpp:247-251`), deliberately: they are decodable for display/history but are no longer valid payment destinations.

`DecodeDestination` still decodes those addresses successfully and returns the destination with `error_str` left empty (`src/key_io.cpp:153-162` and the Dilithium bech32 branch). So `isValid` is false while `error_msg` is empty, and the assertion fires.

## Impact

`validateaddress` is the standard entry point explorers, exchanges, and third-party wallets use to check an address. Every pre-v0.4.2 Dilithium address on testnet triggers it, and the message tells users to report a bug at a dead URL (`github.com/btq/btq`).

Other RPCs are unaffected — `getaddressinfo`, `getdescriptorinfo`, and `deriveaddresses` all return clean `-5` errors for the same input. Only `validateaddress` carries this invariant.

## Fix

Set an error whenever a well-formed address decodes to a destination `IsValidDestination()` rejects, naming P2MR as the replacement so the message is actionable.

augchan42:
Fixes #81.

`validateaddress` asserts `isValid == error_msg.empty()` (`rpc/output_script.cpp:64`). Making Dilithium P2MR-only turned every legacy Dilithium destination invalid, but `DecodeDestination` still decodes them and left `error_str` empty, so the assert fired:

```
$ btq-cli validateaddress nEw8yjgSEg5hqRB6bgkbuQaajkzLc76BN4
error code: -1
Internal bug detected: isValid == error_msg.empty()
```

## Change

`DecodeDestination` now sets an error whenever a well-formed address decodes to something `IsValidDestination()` rejects. Legacy Dilithium addresses get a message naming P2MR as the replacement; anything else gets a generic one, so the invariant holds regardless of which destination types are added later.

Done as a wrapper around the existing decoder rather than at the four return sites, so it cannot drift.

Behaviour after:

```
$ btq-cli validateaddress nEw8yjgSEg5hqRB6bgkbuQaajkzLc76BN4
{
  "isvalid": false,
  "error_locations": [],
  "error": "Legacy Dilithium address: Dilithium is only valid inside P2MR (BIP360) tapscript, so this is not a spendable payment destination. Use getnewdilithiumaddress to obtain a P2MR address."
}
```

No consensus, policy, wallet, or encoding changes. Legacy addresses stay decodable, so history and display are unaffected.

## Tests

New `key_io_tests/key_io_legacy_dilithium_reports_error` locks the invariant across all three decodable legacy forms (base58 pubkey-hash, base58 script-hash, bech32 witness-v0 key-hash), plus a P2MR control asserting it still decodes valid with no error — so the test cannot pass by rejecting everything.

Verified manually on regtest for all four legacy forms, a garbage address, a valid bech32 ECDSA address, and a P2MR address.

`key_io_tests` reports 160 failures both with and without this change — the pre-existing inherited-vector failures tracked in #60. No new failures. `dilithium_address_script_tests` passes clean.

### DIFF
--- a/src/key_io.cpp
+++ b/src/key_io.cpp
@@ -122,7 +122,7 @@ public:
     std::string operator()(const DilithiumPubKeyDestination& pk) const { return {}; }
 };
 
-CTxDestination DecodeDestination(const std::string& str, const CChainParams& params, std::string& error_str, std::vector<int>* error_locations)
+CTxDestination DecodeDestinationInner(const std::string& str, const CChainParams& params, std::string& error_str, std::vector<int>* error_locations)
 {
     std::vector<unsigned char> data;
     uint160 hash;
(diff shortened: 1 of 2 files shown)


Oskii:
Superseded for the Miningcore/testnet Base58 case by #88.

#82 correctly stops the assert by returning `isvalid: false`, but Miningcore requires `isvalid: true` for the live pool coinbase Dilithium Base58 address while `DEPLOYMENT_DILITHIUM_P2MR` is still unscheduled on testnet. #88 restores those destinations on unscheduled networks and keeps the explanatory `isvalid: false` path for witness-v0 / post-activation cases.

augchan42:
Agreed, closing — #88 is the better fix.

The Miningcore constraint is the part I had no visibility into. My change would have returned `isvalid: false` for the pool coinbase address on testnet and broken payouts, which is a strictly worse outcome than the assert it was fixing. Height-gating validity on whether `DEPLOYMENT_DILITHIUM_P2MR` is scheduled is the right call, and it is the same pattern already used for `nDilithiumP2MRHeight` elsewhere.

Worth noting for anyone reading later: landing this on top of #88 would now be actively harmful, not merely redundant. #88 sets the same error strings in `validateaddress`; my version sets them in `DecodeDestination`, so the RPC would find `error_msg` already non-empty and skip its own branch. Two copies of the same message, one of them dead. One place is correct.

The only thing lost is scope — mine fixed the invariant at the source, so any future `DecodeDestination` caller relying on "invalid implies non-empty error" would be safe, whereas #88 fixes it at the single caller that has the assert. I checked earlier that `getaddressinfo`, `getdescriptorinfo` and `deriveaddresses` all return clean `-5` errors, so there is no live gap. Just worth knowing the invariant is enforced per-caller rather than centrally if another one is added.

Thanks for catching the pool dependency.

PR #83, opened 2026-07-27
wallet: sign P2MR inputs in FillPSBT so walletprocesspsbt completes

### LINKED ISSUE #79: psbt: P2MR (Dilithium) inputs are silently skipped during signing — walletprocesspsbt returns complete=false while signrawtransactionwithwallet signs the same inputs fine
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behaviour

Issue analysis with Claude Fable 5.

## Summary. 
On v0.4.2-testnet, walletcreatefundedpsbt + walletprocesspsbt cannot produce a complete PSBT when coin selection includes a P2MR (witness v2) input. The wallet finalizes all Taproot/SegWit inputs and silently skips the P2MR input, returning complete: false with no per-input diagnostics. The same wallet signs the same UTXO set successfully via signrawtransactionwithwallet (complete: true, transaction relays), so the signing capability exists — the gap is PSBT-specific.

## Evidence. 
PSBT with 3 inputs — input 0 witness_v2_p2mr (5220 36223b...), inputs 1–2 witness_v1_taproot. After walletprocesspsbt, inputs 1–2 carry PSBT_IN_FINAL_SCRIPTWITNESS; input 0 carries only witness_utxo. analyzepsbt reports next: "updater" with has_utxo: true for every input, which gives the user no clue what is missing. Full log with PSBT base64: [log at gist](https://gist.github.com/unkn0wn747/9ada4d29e05691318d48f3b3fcb54423).

## Likely root cause. 
The PSBT format carries no fields for P2MR spend data (merkle path / leaf script / control-block analogue, Dilithium key material). The raw-signing path reads these from the wallet's internal P2MR metadata store; the PSBT signer only consults PSBT input fields, finds nothing signable, and skips.

## Suggested direction. 
(1) Define proprietary PSBT key types (or adopt BIP-360's PSBT fields if/when specified) for P2MR spend data; populate them in walletcreatefundedpsbt, consume them in the PSBT signer. (2) Until then, make the failure loud: walletprocesspsbt should report a per-input error (e.g. "P2MR inputs are not supported in PSBT; use signrawtransactionwithwallet"), and analyzepsbt should name the missing data. (3) Consider having PSBT-oriented coin selection avoid P2MR inputs when other funds suffice, since selecting an unsignable-in-this-workflow input guarantees failure.

## Impact. 
Hardware wallets, offline signing, and any PSBT-based coordination silently fail once a wallet holds quantum-safe coins — the flagship output type of the chain. Also mainnet-relevant for exchanges/custodians that are PSBT-based.

## Workarounds. 
Use signrawtransactionwithwallet / the dedicated P2MR RPCs; or pass explicit non-P2MR inputs to walletcreatefundedpsbt.

### Expected behaviour

walletprocesspsbt should either (a) sign P2MR inputs and return complete: true — matching the behaviour of signrawtransactionwithwallet, which successfully signs the exact same inputs from the same wallet — or, if PSBT support for P2MR is not yet implemented, (b) fail loudly: return a per-input error identifying the P2MR input as unsupported in the PSBT workflow (e.g. "error": "P2MR inputs cannot be signed via PSBT; use signrawtransactionwithwallet"), and analyzepsbt should report the specific missing data for that input rather than a generic next: "updater" with has_utxo: true.

Silently returning complete: false with no diagnostics is the worst of both options: the caller cannot distinguish "missing key material", "missing UTXO data", and "unsupported input type", and PSBT-based tooling (hardware wallets, offline signers, coordinators) has no way to detect that quantum-safe coins are unusable in this workflow.

### Steps to reproduce

Tested on v0.4.2-testnet, descriptor wallet (sqlite).

Placeholders: WALLET = wallet name, P2MR_ADDR = address from getnewdilithiumaddress,
P2MR_TXID / VOUT_N = the P2MR outpoint, DEST_ADDR = any destination, PSBT = base64 string.

1. Have a wallet containing at least one confirmed P2MR UTXO (mine or receive to an
   address from `getnewdilithiumaddress`) alongside ordinary ECDSA UTXOs.

2. Create a funded PSBT, forcing the P2MR input so the repro is deterministic:

```bash
btq-cli -testnet -rpcwallet=WALLET listunspent 1 9999999 '["P2MR_ADDR"]'

btq-cli -testnet -rpcwallet=WALLET walletcreatefundedpsbt \
  '[{"txid":"P2MR_TXID","vout":VOUT_N}]' \
  '{"DEST_ADDR":0.4}'
```

3. Process the PSBT:

```bash
btq-cli -testnet -rpcwallet=WALLET walletprocesspsbt "PSBT"
```

Observed: `"complete": false`, no error field. `decodepsbt` on the returned PSBT shows the
P2MR input still carrying only `witness_utxo` (no signature, no final scriptwitness); in a
mixed-input PSBT all Taproot/SegWit inputs do receive PSBT_IN_FINAL_SCRIPTWITNESS, only the
P2MR input is skipped.

4. Analyze gives no actionable information:

```bash
btq-cli -testnet -rpcwallet=WALLET analyzepsbt "PSBT"
# -> every input: {"has_utxo": true, "is_final": false, "next": "updater"}
```

5. Control experiment proving the wallet CAN sign these inputs — same spend via the raw path:

```bash
btq-cli -testnet -rpcwallet=WALLET createrawtransaction \
  '[{"txid":"P2MR_TXID","vout":VOUT_N}]' '{"DEST_ADDR":0.4}'

btq-cli -testnet -rpcwallet=WALLET signrawtransactionwithwallet "RAW_HEX"
# -> "complete": true; the resulting hex passes testmempoolaccept
```

### Relevant log output

FULL VALIDATION LOG IS HERE: https://gist.github.com/unkn0wn747/9ada4d29e05691318d48f3b3fcb54423

### How did you obtain BTQ Core

Compiled from source

### What version of BTQ Core are you using?

v0.4.2

### Operating system and version

Ubuntu 24

### Machine specifications

_No response_

augchan42:
Fixes #79.

`walletprocesspsbt` left P2MR (Dilithium) inputs untouched and returned `complete: false` with an empty `errors` list, while `signrawtransactionwithwallet` signed the same outpoint fine.

## Root cause

This is narrower than the issue thread concluded, and does **not** require new PSBT fields.

Tracked P2MR scriptPubKeys are owned via wallet metadata, not by any `ScriptPubKeyMan`. `CWallet::GetSolvingProvider` already has an explicit fallback for exactly this (`wallet.cpp:3535-3543` — looks up `GetP2MRByScript`, returns `BuildP2MRSigningProvider`), which is why the raw signing path works.

`CWallet::FillPSBT` has no equivalent. It only loops `GetAllScriptPubKeyMans()`, so no SPKM claims a P2MR script, `DescriptorScriptPubKeyMan::FillPSBT` gets a null provider, falls through to ECDSA/Taproot pubkey scavenging, finds nothing, and leaves the input alone without recording an error.

## Change

Add the P2MR pass to `FillPSBT`, mirroring the fallback that already exists in `GetSolvingProvider`. Because the wallet holds the P2MR spend data locally, the same-wallet workflow signs from metadata exactly as the raw path does.

Carrying P2MR spend data *across devices* in a PSBT — hardware wallets, air-gapped signers — is a genuinely separate problem that does need new key types, and is still unsolved. This PR does not address that, and I would not want it to look like it does.

Respects `sign` and `bip32derivs` via `HidingSigningProvider`, and returns `SIGHASH_MISMATCH` on a conflicting per-input sighash type, consistent with the SPKM path.

## Tests

New `wallet_dilithium_psbt.py`:

- P2MR-only PSBT through `walletprocesspsbt` → `finalizepsbt` → `testmempoolaccept` → broadcast → confirmed, asserting the P2MR outpoint was the input actually spent. Mining it proves the Dilithium witness verifies, not just that the PSBT reported completion.
- Mixed P2MR + ordinary input PSBT.
- `sign=false` leaves it unsigned without erroring, and signing the same PSBT afterwards still works.
- Ordinary-only PSBT control.
- Asserts the P2MR scriptPubKey really is witness v2 (`5220…`), so a change to the receive type could not make the test pass vacuously.

**Verified to fail on baseline** at `assert_equal(processed["complete"], True)` → `not(False == True)`, and pass with the fix.

`wallet_dilithium_send.py` passes. `wallet_basic.py` and `rpc_psbt.py` fail identically with and without this change (pre-existing inherited-vector failures), so no new failures.

## Note on analyzepsbt

`analyzepsbt` on an unsigned P2MR PSBT still reports `next: "updater"` rather than naming what is missing. That is now cosmetic — signing works — but it is the remaining half of the "fail loudly" request in the issue, and is worth a follow-up.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2208,6 +2208,41 @@ TransactionError CWallet::FillPSBT(PartiallySignedTransaction& psbtx, bool& comp
         }
     }
 
+    // Tracked P2MR scriptPubKeys are owned via wallet metadata rather than by a
+    // ScriptPubKeyMan, so the loop above never sees them and would leave the
+    // input untouched — complete=false with no indication of which input failed.
+    // GetSolvingProvider() has the same fallback for the raw signing path; this
+    // is what keeps walletprocesspsbt in step with signrawtransactionwithwallet.
+    for (unsigned int i = 0; i < psbtx.tx->vin.size(); ++i) {
+        PSBTInput& input = psbtx.inputs.at(i);
+        if (PSBTInputSigned(input)) continue;
+
+        const CScript* script{nullptr};
+        if (!input.witness_utxo.IsNull()) {
+            script = &input.witness_utxo.scriptPubKey;
+        } else if (input.non_witness_utxo) {
+            const uint32_t n{psbtx.tx->vin[i].prevout.n};
+            if (n >= input.non_witness_utxo->vout.size()) continue;
+            script = &input.non_witness_utxo->vout[n].scriptPubKey;
+        } else {
+            continue;
+        }
+
+        const auto entry{GetP2MRByScript(*this, *script)};
+        if (!entry) continue;
+
+        if (sign && input.sighash_type && *input.sighash_type != sighash_type) {
+            return TransactionError::SIGHASH_MISMATCH;
+        }
+
+        const FlatSigningProvider provider{BuildP2MRSigningProvider(*this, entry->id)};
+        if (SignPSBTInput(HidingSigningProvider(&provider, /*hide_secret=*/!sign, /*hide_origin=*/!bip32derivs),
+                          psbtx, i, &txdata, sighash_type, /*out_sigdata=*/nullptr, finalize) &&
+            n_signed) {
+            (*n_signed)++;
+        }
+    }
+
     RemoveUnnecessaryTransactions(psbtx, sighash_type);
 
     // Complete if every input is now signed
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -128,6 +128,7 @@ BASE_SCRIPTS = [
     'wallet_fundrawtransaction.py --legacy-wallet',
     'wallet_fundrawtransaction.py --descriptors',
     'wallet_dilithium_send.py --descriptors',
+    'wallet_dilithium_psbt.py --descriptors',
     'wallet_bumpfee.py --legacy-wallet',
     'wallet_bumpfee.py --descriptors',
     'wallet_import_rescan.py --legacy-wallet',
--- a/test/functional/wallet_dilithium_psbt.py
+++ b/test/functional/wallet_dilithium_psbt.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 The BTQ Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Regression test for signing P2MR (Dilithium) inputs through the PSBT workflow.
+
+Issue #79: walletprocesspsbt silently skipped P2MR inputs and returned
+complete=false with no diagnostic, while signrawtransactionwithwallet signed the
+same outpoint fine. P2MR scriptPubKeys are tracked in wallet metadata rather than
+by a ScriptPubKeyMan, and CWallet::FillPSBT only consulted ScriptPubKeyMans.
+"""
+
+from decimal import Decimal
+
+from test_framework.test_framework import BTQTestFramework
+from test_framework.util import assert_equal
+
+
+class WalletDilithiumPSBTTest(BTQTestFramework):
+    def add_options(self, parser):
+        self.add_wallet_options(parser, descriptors=True, legacy=False)
+
+    def set_test_params(self):
+        self.num_nodes = 1
+        self.setup_clean_chain = True
+
+    def skip_test_if_missing_module(self):
+        self.skip_if_no_wallet()
+
+    def dilithium_address(self, wallet):
+        created = wallet.getnewdilithiumaddress()
+        assert isinstance(created, dict), created
+        return created["address"]
+
+    def run_test(self):
+        node = self.nodes[0]
+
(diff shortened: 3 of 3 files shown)


augchan42:
Marking draft deliberately — this is **not** incomplete, it is queued.

There are four PRs from me open at once (#82, #83, #84, #87), which is more than anyone should have to absorb in one sitting. #87 and #82 are ten-minute merges and I would rather they land first: #87 restores the ability to trust the test suite at all, and #82 fixes a bug that explorers and exchanges are hitting today.

This one is the most valuable of the four but also the one that deserves real attention — it touches the wallet signing path. Better arriving when someone has time to spend on it than as fourth in a stack.

Will mark ready once #87 and #82 are merged. Undraft sooner if you would rather take it now.

augchan42:
Marking ready — #82 was closed as superseded by #88, so this takes its place in the queue alongside #87.

Rebase check: this touches `src/wallet/wallet.cpp` only, so it does not overlap #88 (`addresstype.cpp`, `output_script.cpp`, `key_io_tests.cpp`) or #89 (`address_all_chains_tests.cpp`, `Makefile.test.include`). GitHub reports it mergeable against current master.

Note on CI: `build-and-address-tests` is failing here, but it fails on master too — three consecutive runs including the #88 and #89 merges. Not introduced by this PR. Tracked in BTQ-90.

PR #84, opened 2026-07-27
rpc: add verifymessagewithdilithium, deprecate verifydilithiumsignature

### LINKED ISSUE #80: rpc: verifydilithiumsignature rejects the P2MR receive address that getnewdilithiumaddress produced (sign works, verify does not)
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behaviour

On master post-#78 (testnet 148777): signmessagewithdilithium <addr> <msg> succeeds for an address freshly returned by getnewdilithiumaddress (produces a ~3200-char signature), but verifydilithiumsignature <same_addr> <sig> <msg> fails with -5 "Address is not a Dilithium key address (use a Dilithium P2MR receive address, or a historical DilithiumPKHash)" — despite that address being exactly a Dilithium P2MR receive address from getnewdilithiumaddress. So sign and verify disagree about what a valid Dilithium address is. Either the verify-side address→key lookup doesn't handle the P2MR form, or it expects a raw pubkey argument the sign side never returns (in which case there is nothing to verify against, and the help text is misleading).

### Expected behaviour

both sign and verify should work

### Steps to reproduce

getnewdilithiumaddress → signmessagewithdilithium → verifydilithiumsignature with the same address.

### Relevant log output

log on latest master as of now 

https://gist.github.com/unkn0wn747/85aa8f31ae8f32cdf29db27ffe8e15df

### How did you obtain BTQ Core

Compiled from source

### What version of BTQ Core are you using?

v0.4.2 master@984264c

### Operating system and version

ubuntu 24

### Machine specifications

_No response_

augchan42:
Closes #80 — which turned out not to be a bug, but pointed at a real inconsistency.

## The report

#80 said `signmessagewithdilithium` succeeds but `verifydilithiumsignature` rejects the same address. It does not — the call was made in the wrong argument order. `verifydilithiumsignature "message" "address" "signature"` returns `true` for an address from `getnewdilithiumaddress`.

## Why it caught an experienced user

Every other Dilithium RPC follows its inherited counterpart. This one does not:

| RPC | order | matches counterpart |
|---|---|---|
| `getnewdilithiumaddress(label, address_type)` | vs `getnewaddress` | yes |
| `importdilithiumkey(privkey, label, rescan)` | vs `importprivkey` | yes |
| `signmessagewithdilithium(address, message)` | vs `signmessage` | yes |
| `signtransactionwithdilithium(hexstring, prevtxs, …)` | vs `signrawtransactionwithwallet` | yes |
| `verifydilithiumsignature(message, address, signature)` | vs `verifymessage(address, signature, message)` | **no** |

Four out of five, so this reads as a slip rather than a house style. Anyone porting `verifymessage` code hits it, and the resulting error — "Address is not a Dilithium key address" — points away from the actual problem.

## Change

- Add `verifymessagewithdilithium(address, signature, message)`, mirroring `verifymessage` in both name and argument order.
- Keep `verifydilithiumsignature` working, marked DEPRECATED in its help, pointing at the new name. **No existing caller breaks.**
- Both delegate to one shared verification body; the only difference is the parameter mapping.
- Both errors now name their own expected argument order, so a mis-ordered call explains itself.

Suggested removal of the old name at the mainnet release, once the deprecation has had a cycle.

## On safety of the reorder

A mis-ordered call cannot silently verify. The message lands where an address is expected and fails to decode; for it to get further, a single string would have to be both a valid Dilithium address and a valid ~3200-char base64 signature. It fails closed, never open — which is what makes introducing a second order acceptable at all. There is a test asserting exactly this.

## Tests

Extends `wallet_dilithium_send.py`: new RPC verifies a good signature, rejects a tampered message, and rejects the deprecated argument order with `-5` rather than returning a result. The existing `verifydilithiumsignature` assertion stays, covering the deprecated path.

`wallet_dilithium_send.py` passes. `rpc_help.py` fails identically with and without this change (pre-existing client-conversion-table mismatch), so no new failures.

## Unrelated observation

While reading the surrounding RPCs: `signtransactionwithdilithium` documents its `prevtxs` argument as literally `"Ignored"`. It accepts the parameter and discards it, which is the same class of footgun. Not touched here — flagging it for a separate look.


augchan42:
Marking draft deliberately — queued, not incomplete.

Four PRs from me are open at once (#82, #83, #84, #87). #87 and #82 are ten-minute merges that I would rather land first. This one is purely additive — a new RPC plus a deprecation notice on the old name — so nothing breaks while it waits, and it is the least urgent of the four.

Will mark ready once the others are through.

BarneyChambers:
Reviewed this for security issues alongside #83 and #87, and built and ran it locally. Leaving it as a draft; notes below.

## Review outcome

No medium-or-higher vulnerability introduced by this PR. The refactor is behaviour-preserving: `DilithiumVerifyMessage` is a faithful extraction of the old `verifydilithiumsignature` body, and the two RPCs differ only in how they map positional arguments onto it.

Specifically checked, and fine:

- **Address-to-key binding.** Verification does not simply trust the supplied address. For P2MR receive addresses the key ID comes from `GetSingleDilithiumKeyIDForP2MR()`, i.e. the key committed to by the address's merkle tree; for the historical `DilithiumPKHash` / `DilithiumWitnessV0KeyHash` forms the key ID is the pubkey hash itself. There's no path where a signature from an unrelated key verifies against an address.
- **Argument-order confusion.** This is the actual risk of shipping two RPCs with permuted signatures, and it's handled well. A mis-ordered call hits the `else` branch and throws `RPC_INVALID_ADDRESS_OR_KEY` rather than silently verifying, and the `usage` string means the error tells the caller the order it expected. The test asserting the deprecated order raises `-5` against the new RPC is the right test to have written.
- **Failure modes.** Bad base64 and invalid pubkeys throw rather than returning a bare `false`, so a caller cannot confuse "malformed input" with "signature did not verify".

## Validation

- Merges cleanly into `master`, builds clean.
- `wallet_dilithium_send.py --descriptors` passes, including the new assertions.
- The other functional failures I saw (`rpc_psbt.py`, `rpc_signmessagewithprivkey.py`) reproduce identically on unmodified `master`, so they are pre-existing and unrelated to this PR. `rpc_signmessagewithprivkey.py` in particular fails on a hardcoded expected-signature vector inherited from upstream, not on anything this PR touches.

## One thing to be aware of, not caused by this PR

While tracing the verification path I found that Dilithium message signing has no domain separation from transaction signing: `CDilithiumKey::Sign(hash)` is literally `SignMessage(hash_bytes, empty_context)`, and `signmessagewithdilithium` signs raw user bytes through the same call with the same empty context. Since a P2MR address's message-signing key is the same key that spends the output, a signature over a 32-byte "message" is byte-identical to a spending signature.

That's pre-existing on `master` and this PR neither introduces nor worsens it, so it isn't a blocker here. Filed separately as #90.

It is, however, worth noting for the deprecation plan: whichever way #90 is fixed will change the signature format, so `verifydilithiumsignature` and `verifymessagewithdilithium` have to move in lockstep. It may be cheaper to land the domain-separation change and this RPC rename together, so callers migrate once instead of twice.

BarneyChambers:
Merged master in to resolve a conflict in `src/wallet/rpc/dilithium.cpp`.

The conflict was substantive rather than textual. This PR factored the verification body out into the shared `DilithiumVerifyMessage` helper, and in parallel #126 rewrote that same body in place to verify against a separate message-signing domain, so a transaction signature made with the same key no longer verifies as a message signature.

Taking either side wholesale would have been wrong: keeping this branch's helper as written would have reinstated the raw `dilithium_pubkey.VerifyMessage(...)` call and silently dropped #126's domain separation. The helper now ends in `DilithiumMessageVerify(dilithium_pubkey, strMessage, strSignature)` instead, so both `verifydilithiumsignature` and the new `verifymessagewithdilithium` inherit the domain check. The deprecation note and #126's note about domains are both kept in the help text.

Builds clean, and the 90 Dilithium unit tests pass.

PR #88, opened 2026-07-27
fix: keep testnet legacy Dilithium Base58 addresses valid

### LINKED ISSUE #81: rpc: validateaddress returns "Internal bug detected" for any legacy Dilithium address

Oskii:
## Summary
- Restore `DilithiumPKHash` / `DilithiumScriptHash` as valid payment destinations while `DEPLOYMENT_DILITHIUM_P2MR` remains unscheduled (testnet today), so Miningcore `validateaddress` accepts the pool coinbase address again.
- On chains with P2MR-only from genesis/height 1, those destinations stay invalid; `validateaddress` now returns `isvalid: false` with an explanatory error instead of tripping `isValid == error_msg.empty()`.
- Adds a unit test covering the live mining-pool Dilithium Base58 address on testnet.

Supersedes the `isvalid: false` approach in #82 for testnet Base58 Dilithium (still needed there for witness-v0 / post-activation cases). Fixes #81 without breaking Miningcore.

## Test plan
- [x] `test_btq --run_test=key_io_tests/key_io_testnet_legacy_dilithium_base58_still_valid`
- [x] `test_btq --run_test=dilithium_p2mr_only_tests/legacy_dilithium_destinations_not_valid_payments`
- [x] `btq-cli -testnet validateaddress nSUTw4YKuND8kNFad9TyWBXCKouru85oGa` → `isvalid: true`, `isdilithium: true`
- [x] `btq-cli -regtest validateaddress <legacy Dilithium>` → `isvalid: false` with error (no assert)
- [ ] Upgrade mining pool node and confirm Miningcore stays up on :3333

### DIFF
--- a/src/addresstype.cpp
+++ b/src/addresstype.cpp
@@ -4,6 +4,7 @@
 
 #include <addresstype.h>
 
+#include <chainparams.h>
 #include <crypto/dilithium_key.h>
 #include <crypto/sha256.h>
 #include <hash.h>
@@ -14,6 +15,7 @@
 #include <util/hash_type.h>
 
 #include <cassert>
+#include <limits>
 #include <vector>
 
 typedef std::vector<unsigned char> valtype;
@@ -229,6 +231,15 @@ public:
     }
 };
 
+/** True while DEPLOYMENT_DILITHIUM_P2MR remains unscheduled (testnet today).
+ *  Legacy Dilithium Base58 P2PKH/P2SH outputs are still consensus-spendable
+ *  there; on chains that activate P2MR-only from genesis/height 1 they are not
+ *  valid payment destinations. */
+bool LegacyDilithiumBase58PaymentsAllowed()
+{
+    return Params().GetConsensus().nDilithiumP2MRHeight == std::numeric_limits<int>::max();
+}
+
 class ValidDestinationVisitor
 {
 public:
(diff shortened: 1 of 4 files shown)


Oskii (commented):
Looks correct and safe to land.

`IsValidDestination` again treats Dilithium Base58 P2PKH/P2SH as valid payment destinations only while `nDilithiumP2MRHeight == INT_MAX` (testnet today). Mainnet/regtest/signet stay P2MR-only from height 1, so those destinations remain invalid there. Witness-v0 Dilithium and bare Dilithium P2PK stay invalid everywhere. The `validateaddress` error fill preserves the assert invariant for well-formed-but-invalid destinations.

Locally verified: `key_io_testnet_legacy_dilithium_base58_still_valid` and `dilithium_p2mr_only_tests` pass, including the live pool address `nSUTw4YKuND8kNFad9TyWBXCKouru85oGa`.

Lean CI `make check` is red on `bip324_tests` (pre-existing on master; unrelated to this change). Merging so the mining-pool fix is on master rather than only an ad-hoc binary.

PR #91, opened 2026-07-30
test: unblock the unit suite — fix descriptor_tests segfault and the failures it was hiding

BarneyChambers:
## Summary

The unit test binary was segfaulting in `descriptor_tests`, the 37th of 105 suites. Because Boost runs suites in one process, everything ordered after it never executed 

This fixes that crash and the four other failures it was masking. Five suites go from failing to green (`descriptor_tests`, `bip324_tests`, `denialofservice_tests`, `key_tests`, `key_io_tests`), and the run now reaches **105 suites instead of 44**.

Only test files change. No production code is touched.

## Root cause

Almost everything here is one pattern: test data inherited from upstream Bitcoin that BTQ's changed constants invalidated. PR #87 fixed the first instance of it in `bloom_tests`; these are the rest.

| Constant | Bitcoin | BTQ |
|---|---|---|
| `PUBKEY_ADDRESS` / `SCRIPT_ADDRESS` / `SECRET_KEY` (mainnet) | 0 / 5 / 128 | 75 / 135 / 235 |
| `EXT_PUBLIC_KEY` / `EXT_SECRET_KEY` (mainnet) | `0488B21E` / `0488ADE4` | `0488B21F` / `0488ADE5` |
| bech32 HRP (main/test/signet/regtest) | `bc` / `tb` / `tb` / `bcrt` | `qbtc` / `tbtq` / `qtb` / `qcrt` |
| BIP324 HKDF salt | `bitcoin_v2_shared_secret` + `f9beb4d9` | `btq_v2_shared_secret` + `f1b2a3d4` |
| `MAX_SCRIPT_ELEMENT_SIZE` | 520 | 15000 |
| `nPowTargetSpacing` | 600s | 60s |

## What changed

**`descriptor_tests`** — the crash. Bitcoin-mainnet WIF/`xprv`/`xpub` literals no longer decode, `Parse()` correctly rejected them, and the test dereferenced the resulting null `unique_ptr`. Re-encoded 49 literals (372 occurrences) under BTQ's prefixes with the key bytes unchanged, recomputed the 18 descriptor checksums that invalidated (preserving each `CheckUnparsable` case's intent — valid, corrupted, too long, too short), regenerated the 13 COMPAT `DescriptorID` values, and promoted the two post-`Parse()` checks to `BOOST_REQUIRE` so this can never take down the binary again.

Two cases asserting a 16-key P2SH redeemScript is oversized now assert it parses instead: BTQ raised `MAX_SCRIPT_ELEMENT_SIZE` to 15000 for Dilithium (keys 1312–2592 bytes, signatures 2420–4627), which makes that limit unreachable for `multi()` at its 20-key cap.

**`bip324_tests`** — 280 failures. BTQ changed both halves of the BIP324 HKDF salt, so every session id, garbage terminator and ciphertext differs. The transport is faithful to upstream; only the vectors were stale.

**`key_tests`** — the same stale WIFs as `bloom_tests`. `GetPubKey()` tripped `assert(keydata)`, the aborted fixture left the logger mid-init so the next case died on `assert(m_buffering)`, and the binary then hung.

**`key_io_valid.json`** — 42 of 70 vectors re-encoded for BTQ's mainnet base58 versions and all four bech32 HRPs. Payload hex and metadata untouched.

**`denialofservice_tests`** — not a data problem. The mocked clock jump is `3 * nPowTargetSpacing + 1s`, but `CheckForStaleTipAndEvictPeers()` only inspects the tip once every `STALE_CHECK_INTERVAL` (a fixed 10 minutes). Upstream's 10-minute blocks clear both thresholds; BTQ's 1-minute blocks make the jump 181s, so the check never ran.

## On regenerating vectors

Where values had to be regenerated, they were derived independently and the derivation was proven first, rather than recording whatever the implementation emits:

- **BIP324**: rebuilt with upstream Bitcoin Core's own Python reference (`test_framework/crypto/{chacha20,poly1305,bip324_cipher,hkdf,ellswift}.py`). The generator was first run with *upstream's* salt and reproduced all seven original vectors exactly; only then was it used to emit BTQ's. Recording the C++ output would have left 280 assertions that can never fail.
- **`key_io` vectors**: the converter was first run with Bitcoin's parameters and required to reproduce all 70 originals byte for byte.
- **bech32 address in `descriptor_tests`**: reproduces the old `bc1q…` value exactly under the `bc` HRP.
- **P2SH scriptPubKeys**: built from the pubkeys in the descriptor. Worth noting the 16-of-17 redeemScript is 582 bytes, not the 581 the old error message quoted — `descriptor.cpp` estimates size as `sum(pubkey sizes) + 3`, which assumes single-opcode pushes for m and n, but 17 has no opcode and needs a pushdata.

The COMPAT `DescriptorID` values are the one exception: they are `SHA256(desc.ToString(compat_format=true))` and were regenerated from the implementation, since they had to change once the key material was re-encoded. That was explicitly blessed rather than done silently.

## What this does not fix

CI will still be red, but for different and further-along reasons. With the binary no longer dying at `descriptor_tests`, 61 additional suites run for the first time, and **22 pre-existing failing cases plus a hang are now visible**. None are caused by this PR — it only changes test files, and all seven suites it touches report zero failures.

| Suite | Cases | Notes |
|---|---|---|
| `miner_tests/CreateNewBlock_validity` | 112 + abort | largest single group |
| `txpackage_tests` | 4 cases, 21 assertions, aborts | |
| `validation_chainstatemanager_tests` | 5 cases, 9 assertions, aborts | |
| `script_standard_tests/bip341_spk_test_vectors` | 7 | likely stale taproot vectors |
| `util_tests/message_verify`, `message_sign` | 7 | stale mainnet addresses |
| `miniscript_tests/fixed_tests` | 5 | |
| `validation_tests/test_assumeutxo` | 4 | |
| `rpc_tests/rpc_rawsign`, `psbt_wallet_tests` | 3, aborts | stale WIFs / `xprv` |
| `script_parse_tests`, `util_tests/test_Capitalize`, `validation_chainstate_tests` | 3 | |
| `spend_tests/SubtractFee` | **hangs** | binary stops here |

Some of these look like more of the same stale-data pattern (`util_tests` message signing, `rpc_tests`, `psbt_wallet_tests`, and probably the BIP341 vectors). The `miner_tests`, `txpackage_tests` and `validation_*` groups look different in character and may be real issues with the fork rather than test staleness — they deserve a proper look rather than being assumed stale.

## Test plan

- [x] `descriptor_tests`, `bip324_tests`, `denialofservice_tests`, `key_tests`, `key_io_tests` all pass
- [x] `bloom_tests` and `bip32_tests` still pass (deliberately left alone — `bloom_tests`' remaining match is inside PR #87's comment, and `bip32_tests` mixes BTQ-encoded vectors with Bitcoin-encoded ones on purpose)
- [x] No production code changed
- [x] BIP324 generator reproduces all 7 upstream vectors under upstream's salt before emitting BTQ's
- [x] `key_io` converter reproduces all 70 originals under Bitcoin's parameters before emitting BTQ's
- [ ] Full suite still red on 22 pre-existing cases listed above — separate work


augchan42:
Checked both of the open decisions against master (`731c7efc5`) rather than just reading the write-up. Notes on each:

**The P2SH-oversize cases — your analysis holds, and I'd argue it's not a test decision at all.**

`MAX_SCRIPT_ELEMENT_SIZE` is 15000 at `src/script/script.h:29`, and `multi()` is capped at `MAX_PUBKEYS_PER_MULTISIG = 20`. So the guard at `descriptor.cpp:1673` — `script_size + 3 > MAX_SCRIPT_ELEMENT_SIZE` — cannot fire for any `multi()` descriptor, exactly as you describe. The two cases are asserting an error path that no input can reach.

Worth flagging alongside it: the comment immediately above that check still reads

```
// This limits the maximum number of compressed pubkeys to 15.
```

which was true at Bitcoin's 520 and is now off by three orders of magnitude. Whoever makes the call on the test cases should probably fix or drop that line in the same change, since it's the thing that would mislead the next reader into thinking the guard still does something.

The broader question your PR is really surfacing — whether a P2SH redeemScript approaching 15000 bytes is intended to be valid and standard on BTQ, and who owns that limit — seems worth settling explicitly before the external audit rather than leaving it implied by a raised constant. Happy to write that up separately if it's useful; I didn't want to file it on your behalf.

**The 24 COMPAT ids — agree with not regenerating them silently.**

The assertion exists to catch exactly the thing you'd be doing, so a deliberate blessing is the right bar. The keys are byte-for-byte identical under the re-encoding and only the version prefix moves, so the ids *must* change; that's a consequence of the fix, not evidence of a defect. Recording that reasoning in the commit message is probably enough to satisfy it.

**On the CI note in your summary — confirmed by measurement.**

`check-local` runs one test file per `--run_test` invocation and `make check` stops at the first failing file. On current master every one of the 280 reported failures is a `test/bip324_tests.cpp` line; nothing else contributes, and no file sorted after it executes at all. So this PR genuinely cannot show green until the bip324 vectors are skipped, and neither can anything else.

I've raised BTQ-113 (skip bip324 with documented rationale) to Urgent on that basis and am picking it up now — it should clear the way for this one to get a real CI result.

augchan42:
Built this locally on top of #92 (which skips the `bip324_tests` vectors, otherwise the run never gets here) and ran the full binary. Independent confirmation of both your claims:

**The segfault is gone and the failure count is exactly what you said.** `descriptor_tests` runs to completion and fails **30** assertions — your 24 COMPAT ids + 6 P2SH-oversize, to the number. I did not have to take the breakdown on trust.

**It moves the suite materially further:**

| | Suites executed | Terminates at | Exit |
| -- | -- | -- | -- |
| master + #92 | 36 | `descriptor_tests` segfault | 139 (SIGSEGV) |
| master + #92 + #91 | **46** | `key_tests` abort | 134 (SIGABRT) |

Ten more suites execute with this applied. `key_io_tests` also now runs and reports its 160 failures, which matches the count on BTQ-90 — worth noting those fail loudly but do *not* abort, so they were never part of the fatal chain.

**The next fatal, for your notes:**

```
test/key_tests.cpp(44): error: in "key_tests/key_test1":
    check key2C.IsValid() && key2C.IsCompressed() has failed
test_btq: key.cpp:189: CPubKey CKey::GetPubKey() const: Assertion `keydata' failed.
```

Same root-cause pattern #87 fixed in `bloom_tests` — inherited Bitcoin WIF fails to decode under BTQ params, invalid `CKey`, `GetPubKey()` trips the assert. Picking that one up next so it does not block you.

On the P2SH-oversize question, I left a note further up confirming the `MAX_SCRIPT_ELEMENT_SIZE` analysis holds and flagging that the comment above `descriptor.cpp:1673` still claims a 15-pubkey limit. Nothing here changes my read that both open items are genuine decisions rather than things you should just regenerate.

PR #92, opened 2026-07-30
test: skip inherited BIP324 vectors that cannot pass under BTQ params

augchan42:
## Summary

`bip324_tests` is the first failing file in the unit run, and it is currently the only thing CI ever gets to. Skipping its inherited vectors lets the merge gate advance for the first time in weeks.

Tracked as BTQ-113, under the test-suite umbrella BTQ-90.

## Why this is the head of the queue

`check-local` invokes `test_btq` once per test file, and `make check` stops at the first failing file. Measured on master `731c7efc5`:

- the failing step reports **280 failures, and every failure line is `test/bip324_tests.cpp`** — nothing else contributes
- `bip324_tests` sorts early, so **no test file after it executes in CI at all**

So `build-and-address-tests` being red is not a weak signal, it is an absent one. Every other suite — including the work in #91 — is invisible to CI until this lands. That also means red CI is currently meaningless as a review signal on any BTQ PR, which is a bad habit to carry into an audit.

## Why skipping is defensible here

Two facts, both verified against master rather than assumed:

1. **The feature is off.** `DEFAULT_V2_TRANSPORT{false}` (`src/net.h:111`). Upstream Bitcoin ships this `true`; BTQ does not enable it. The code under test is inert in every configuration we ship.
2. **The vectors mismatch by construction, not by defect.** `BIP324Cipher::Initialize` salts the session KDF with `"btq_v2_shared_secret"` plus the network magic (`src/bip324.cpp:38`). BTQ changed both, so every hardcoded upstream packet vector must diverge *even if this implementation is byte-perfect*.

This removes the vectors **only**. BIP324 itself is untouched and still compiles.

Worth stating explicitly, because the short version gets misremembered as "BIP324 isn't quantum-safe so we don't need it" — that argument proves far too much, and by the same standard we would drop the ECDSA compatibility paths. BIP324 was never a post-quantum feature. It defends against classical adversaries that exist today: transaction-origin linkage (which node announced first ties a tx to an IP) and DPI censorship, the latter mattering disproportionately for a small chain still bootstrapping its peer graph. The argument for skipping is the maintenance cost of the vectors against a transport we have not enabled — not the feature being pointless.

Parity is deferred, not abandoned. Regenerating them needs an independent reference generator, and BIP324 is in any case the wrong foundation for a PQ transport (its handshake is ElligatorSwift over secp256k1 ECDH), so that work risks being thrown away against a design that would be replaced rather than extended.

## Implementation note — `precondition()`, not `disabled()`

This is the part worth reviewing carefully, because the obvious approach silently does nothing.

`disabled()` is what you would reach for. It does not work here. The `%.cpp.test` rule in `src/Makefile.test.include` greps the suite name out of the source and passes it to `test_btq` as an explicit `-t bip324_tests` filter — and **an explicit filter overrides `disabled()`**, so the suite runs and fails exactly as before. A `precondition` is evaluated at runtime and survives the filter.

Measured on Boost 1.74, matching CI's ubuntu-22.04, under the exact `-t` invocation the Makefile uses:

| Mechanism | Result under `-t bip324_tests` | Exit |
| -- | -- | -- |
| `*boost::unit_test::disabled()` | suite runs, vectors fail | 201 |
| `*boost::unit_test::precondition(...)` | `skipped because precondition failed` | 0 |

The rationale and this gotcha are both recorded in the file so the next person does not rediscover it by shipping a no-op.

## What this does not do

**It does not turn CI green.** It lets CI advance to the next failing file — most likely `key_io_tests` (160 failures, which must be *regenerated*, not skipped, since they cover BTQ's own address encode/decode). That is the intended outcome: the first honest, ordered account of what is actually broken.

## Test plan

- [x] Suite-name extraction still yields `bip324_tests` after the edit — the added comment block could have corrupted the `-t` filter and does not
- [x] Declaration form (fixture + precondition decorator) compiles clean under `-Wall -Wextra` against `<boost/test/unit_test.hpp>`
- [x] Mechanism verified on Boost 1.74 under the exact Makefile invocation (table above)
- [ ] **Full `make check` not run locally** — `libboost_unit_test_framework` is not installed on my machine, so `test_btq` cannot link. CI on this PR is the end-to-end check; expect it to advance past `bip324_tests` and fail on the next file rather than go green.

## Review note

Needs the confirmation BTQ-113 already calls for: that BTQ does not intend to maintain BIP324 v2 vector parity. If parity is wanted, this re-scopes to "regenerate 280 vectors against BTQ transport params" — real work needing a reference generator — and should be re-scoped rather than skipped.

### DIFF
--- a/src/test/bip324_tests.cpp
+++ b/src/test/bip324_tests.cpp
@@ -158,7 +158,44 @@ void TestBIP324PacketVector(
 
 }  // namespace
 
-BOOST_FIXTURE_TEST_SUITE(bip324_tests, BasicTestingSetup)
+// The vectors below are inherited verbatim from upstream Bitcoin and cannot pass
+// on BTQ, so this suite is skipped rather than asserted. In full:
+//
+//  1. The feature is off. v2 transport is opt-in and disabled by default on BTQ
+//     (DEFAULT_V2_TRANSPORT{false}, src/net.h:111) where upstream ships it
+//     enabled. The code under test is inert in every configuration we ship.
+//  2. The vectors mismatch by construction, not by defect. BIP324Cipher salts
+//     the session KDF with "btq_v2_shared_secret" plus the network magic
+//     (src/bip324.cpp:38). BTQ changed both, so every hardcoded upstream packet
+//     vector must diverge even if this implementation is byte-perfect. These
+//     failures are evidence of deliberately changed parameters, not of broken
+//     crypto.
+//  3. Parity is deferred, not abandoned. Regenerating them against BTQ transport
+//     params requires an independent reference generator, and BIP324 is in any
+//     case the wrong foundation for a post-quantum transport -- its handshake is
+//     ElligatorSwift over secp256k1 ECDH. That work risks being thrown away
+//     against a design that would be replaced rather than extended.
+//
+// This skips the test vectors ONLY. It does not remove or disable BIP324 itself,
+// which still compiles and still carries real value against classical adversaries
+// -- transaction-origin linkage and DPI censorship -- whenever BTQ enables it.
+// "Not post-quantum" is deliberately not the argument here; by that standard we
+// would also drop the ECDSA compatibility paths.
+//
+// Tracked in BTQ-113. Revisit if BTQ enables v2 transport or commits to parity.
+//
+// NB: this uses precondition() and not the more obvious disabled(). The
+// %.cpp.test rule in src/Makefile.test.include greps the suite name out of this
+// file and passes it to test_btq as an explicit `-t bip324_tests` filter, and an
+// explicit filter overrides disabled() -- the suite would run and fail anyway.
+// A precondition is evaluated at runtime and survives the filter. Measured on
+// Boost 1.74: disabled() under `-t` exits 201, precondition under `-t` exits 0.
+static bool upstream_vectors_apply_to_btq(boost::unit_test::test_unit_id)
+{
+    return false;
+}
+
+BOOST_FIXTURE_TEST_SUITE(bip324_tests, BasicTestingSetup, *boost::unit_test::precondition(upstream_vectors_apply_to_btq))
 
 BOOST_AUTO_TEST_CASE(packet_test_vectors) {
     // BIP324 key derivation uses network magic in the HKDF process. We use mainnet params here


augchan42:
Closing the open box in the test plan: the suite now builds and runs locally, so this is verified end-to-end rather than by isolated probe.

Built at CI parity (`--disable-bench --disable-fuzz --disable-fuzz-binary --without-gui --with-incompatible-bdb`), then run through the exact invocation `%.cpp.test` uses:

```
$ ./src/test/test_btq --catch_system_errors=no -l test_suite -t "bip324_tests"
*** No errors detected
EXIT=0
```

Was exit 201 with 280 failures before the change. So the merge gate will now get past this file.

Correcting one thing from the PR description: it says a full `make check` was not run locally because `libboost_unit_test_framework` was unavailable. That was wrong — `src/test/main.cpp` uses `<boost/test/included/unit_test.hpp>`, the header-only variant, so the linked Boost.Test library is not a dependency of this project at all and nothing needed installing. Leaving the note here rather than silently editing the description.

Re-running the full suite now to get an ordered account of what the gate hits next; will follow up with the result.

BarneyChambers:
Closing this as superseded by #91, which landed in `cc14794`.

The premise here was that the upstream vectors "mismatch by construction" and that regenerating them "requires an independent reference generator" — both true, and that generator is what #91 built. It uses upstream Bitcoin Core's own Python reference (`test_framework/crypto/{chacha20,poly1305,bip324_cipher,hkdf,ellswift}.py`), and it was first run with *upstream's* salt and required to reproduce all seven original vectors exactly before being pointed at BTQ's `btq_v2_shared_secret` + `f1b2a3d4`. So the vectors stay an independent check on the transport rather than a recording of its output.

Verified on master's tree just now:

```
test/bip324_tests.cpp(161): Entering test suite "bip324_tests"
--- test_btq exit code: 0
```

Merging this on top would apply the precondition to a suite that now passes, silently removing the only coverage the encrypted transport has. Given the external audit, that seems like the wrong direction.

Three things in here are worth keeping, though, and I don't want them lost with the branch:

1. **`DEFAULT_V2_TRANSPORT{false}` (src/net.h:111) diverges from upstream, which ships it enabled.** That's a real and separately interesting fact — it means the transport is inert in every configuration we ship, and it isn't documented anywhere else that I can find. Worth its own issue.
2. **The point that BIP324's handshake is ElligatorSwift over secp256k1 ECDH, and so is the wrong foundation for a post-quantum transport.** That's a design question that outlives these vectors, and it's exactly the kind of thing the audit should be pointed at deliberately rather than stumble across.
3. **The `disabled()` vs `precondition()` finding.** That the `%.cpp.test` rule in `src/Makefile.test.include` greps the suite name and passes it as an explicit `-t` filter, which overrides `disabled()` but not a runtime precondition, is genuinely useful and non-obvious — measured exit 201 vs 0 on Boost 1.74. That belongs in the developer notes; whoever next tries to skip a suite will otherwise rediscover it the hard way.

Happy to file (1) and (2) if you'd rather not, and thanks for the careful review on #91 — the `MAX_SCRIPT_ELEMENT_SIZE` confirmation and the catch on the stale `descriptor.cpp:1673` comment were both good, and I'm picking the latter up.

augchan42:
Closing as superseded by #91, which regenerates these vectors instead of skipping them. That is the better outcome and I am happy to be wrong here.

I argued on BTQ-113 that regenerating was plausibly throwaway work against a transport that is off by default and whose ElligatorSwift-over-secp256k1 handshake is the wrong foundation for a PQ design anyway. The deferral logic was sound but the cost estimate was not — #91 shows the job was a couple of hours, not a project. Skipping would have removed 280 assertions; regenerating keeps them, and they now actually assert something about BTQ.

The method in #91 is also the right one: upstream Bitcoin Core's Python reference implementation rather than this codebase's C++, with the generator first validated by reproducing all seven original vectors under upstream's salt. That is what makes the new vectors an independent check rather than a recording of our own output — the failure mode that would have left 280 assertions incapable of ever failing.

Verified locally at CI parity on #91's head (`d1c4edd62`): `bip324_tests` runs and passes, and the suite reaches 119 suites / 665 cases with 171 failures, down from 362.

One thing worth carrying forward, since it will bite the next suite-level skip in this repo: `%.cpp.test` in `src/Makefile.test.include` greps the suite name out of the source and passes it to `test_btq` as an explicit `-t` filter, and **an explicit filter overrides `boost::unit_test::disabled()`** — the suite runs and fails regardless. Measured on Boost 1.74: `disabled()` under `-t` exits 201, `precondition` exits 0. Recorded on BTQ-90 so nobody rediscovers it by shipping a no-op.

PR #93, opened 2026-07-30
test: re-encode key_tests key material under BTQ base58 versions

augchan42:
## Summary

`key_tests` aborted the whole test binary. This fixes it, and it turns out to be the link in the fatal chain that unblocks most of the suite: with it applied the run goes from 46 executed suites to **118**.

Tracked as BTQ-112 under the test-suite umbrella BTQ-90.

## Root cause

Same pattern as #87 (`bloom_tests`) and #91 (`descriptor_tests`): test data inherited from upstream Bitcoin that BTQ's changed constants invalidate.

The WIFs and addresses at the top of the file are encoded under Bitcoin's base58 versions, and BTQ uses different ones (`src/kernel/chainparams.cpp`):

| | Bitcoin | BTQ |
|---|---|---|
| `PUBKEY_ADDRESS` | 0 | 75 |
| `SECRET_KEY` | 128 | 235 |

So `DecodeSecret()` returned an invalid `CKey`, four assertions failed, and the first `GetPubKey()` then tripped `assert(keydata)` at `key.cpp:189` — SIGABRT, taking every suite ordered after `key_tests` with it.

## What changed

**Re-encoded rather than sidestepped.** Only the version byte moves; the 32 secret bytes and the compression flag are byte-for-byte identical to what the Bitcoin literals encoded. Every downstream expectation in the file — pubkeys, signatures, negation, ellswift — still holds with no other edits. Keeping the values flowing through `DecodeSecret()` also means this suite still exercises BTQ's own WIF decode path instead of bypassing it, which matters given that path had a live bug in #81.

**`strAddressBad` re-encoded as a BTQ address**, not left as the Bitcoin one. Worth a reviewer's eye, as it is the one change not strictly forced by the crash: leaving it would still pass, but only by accident — it would be checking that a *foreign-chain* string is not a private key, when the upstream intent was that a well-formed address of *the same chain* is not one. That is a weaker test wearing the same name.

**Derivation is independent.** Values were produced with `test/functional/test_framework` (base58 + `ECKey` + `ripemd160`), not recorded from the C++ under test. The same pipeline run with *Bitcoin's* versions reproduces all eight original literals exactly — that round-trip is what makes the BTQ ones trustworthy rather than merely self-consistent.

## Result

`key_tests`: **6 cases, 0 failures.** Was 4 failures plus SIGABRT.

Measured locally at CI parity (`--disable-bench --disable-fuzz --disable-fuzz-binary --without-gui --with-incompatible-bdb`), stacking the open fixes:

| Applied | Suites executed | Terminates at | Exit |
|---|---|---|---|
| #92 | 36 | `descriptor_tests` segfault | 139 |
| #92 + #91 | 46 | `key_tests` abort | 134 |
| #92 + #91 + **this** | **118** (664 cases) | `psbt_wallet_tests` abort | 134 |

## What this exposes

This does not make the suite green — it makes it *visible*. Most of these suites had never executed, so these counts are new information rather than regressions:

| Suite | Failures | Previously known? |
|---|---|---|
| `key_io_tests` | 160 | yes — regenerate, do not skip |
| `miner_tests` | 112 | **no** |
| `descriptor_tests` | 30 | yes — #91's two open decisions |
| `txpackage_tests` | 21 | **no** |
| `validation_chainstatemanager_tests` | 10 | **no** |
| `util_tests` | 8 | yes |
| `script_standard_tests` | 8 | **no** |
| `miniscript_tests` | 5 | **no** |
| `validation_tests` | 4 | **no** |
| `validation_chainstate_tests` / `script_parse_tests` / `rpc_tests` / `denialofservice_tests` | 1 each | partly |

362 failures total, and that is still a **lower bound** — the run aborts at `psbt_wallet_tests`, so anything sorted after it remains unexecuted.

## Next fatal, for whoever picks up the chain

```
test_btq: wallet/test/psbt_wallet_tests.cpp:24:
    void wallet::psbt_wallet_tests::import_descriptor(wallet::CWallet&, const string&):
    Assertion `desc' failed.
```

A descriptor string fails to parse and the result is dereferenced unchecked — same shape again, and the same `BOOST_REQUIRE`-instead-of-deref lesson #91 applies to `descriptor_tests`.

## Test plan

- [x] `key_tests` passes in isolation under the exact `%.cpp.test` invocation: `test_btq -t key_tests` → 6 cases, `*** No errors detected`, exit 0
- [x] Confirmed the cases actually execute rather than being skipped
- [x] Derivation pipeline validated by reproducing all eight original Bitcoin literals byte-for-byte before emitting BTQ values
- [x] Full-binary run stacked with #91 and #92: 118 suites / 664 cases, no abort until `psbt_wallet_tests`
- [ ] CI will still be red — expected, and not caused by this PR

### DIFF
--- a/src/test/key_tests.cpp
+++ b/src/test/key_tests.cpp
@@ -18,16 +18,35 @@
 
 #include <boost/test/unit_test.hpp>
 
-static const std::string strSecret1 = "5HxWvvfubhXpYYpS3tJkw6fq9jE9j18THftkZjHHfmFiWtmAbrj";
-static const std::string strSecret2 = "5KC4ejrDjv152FGwP386VD1i2NYc5KkfSMyv1nGy1VGDxGHqVY3";
-static const std::string strSecret1C = "Kwr371tjA9u2rFSMZjTNun2PXXP3WPZu2afRHTcta6KxEUdm1vEw";
-static const std::string strSecret2C = "L3Hq7a8FEQwJkW1M2GNKDW28546Vp5miewcCzSqUD9kCAXrJdS3g";
-static const std::string addr1 = "1QFqqMUD55ZV3PJEJZtaKCsQmjLT6JkjvJ";
-static const std::string addr2 = "1F5y5E5FMc5YzdJtB9hLaUe43GDxEKXENJ";
-static const std::string addr1C = "1NoJrossxPBKfCHuJXT4HadJrXRE9Fxiqs";
-static const std::string addr2C = "1CRj2HyM1CXWzHAXLQtiGLyggNT9WQqsDs";
+// These were inherited from upstream Bitcoin and encoded under Bitcoin's base58
+// versions, so none of them decoded here: BTQ uses SECRET_KEY 235 (not 128) and
+// PUBKEY_ADDRESS 75 (not 0), see src/kernel/chainparams.cpp. DecodeSecret()
+// returned an invalid CKey, and the GetPubKey() calls below then tripped
+// assert(keydata) in key.cpp and took the whole test binary down with SIGABRT --
+// the same failure mode PR #87 fixed in bloom_tests.
+//
+// Re-encoded under BTQ's versions. Only the version byte changes; the 32 secret
+// bytes and the compression flag are byte-for-byte identical to what the Bitcoin
+// literals encoded, so every downstream expectation in this file (pubkeys,
+// signatures, negation, ellswift) still holds unchanged.
+//
+// Derived with test/functional/test_framework (base58 + ECKey + ripemd160), not
+// by recording what the C++ under test emitted. The same pipeline run with
+// Bitcoin's versions reproduces all eight original literals exactly, which is
+// what makes these trustworthy.
+static const std::string strSecret1 = "8sxAojhKnDu3yKzu3uifzQph8w7zHMJ9V4UGfktD2syDKLTwqiJ";
+static const std::string strSecret2 = "8uBiXYsdvSNJT2TQP4Y1YXAa1aSSdfvMdkZS7ostNbyikjv23Jd";
+static const std::string strSecret1C = "bmtWHHSytbvk7crYjqgw6Q9WA1sqqAagSrKNA4h27Hy7MtuKDJXU";
+static const std::string strSecret2C = "bsLJHqgVxry21sRYCNbsQ89EhYbJ8rnW5DG9s3ubkMPMHxC75WoV";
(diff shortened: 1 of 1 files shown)


BarneyChambers:
Closing this as superseded by #91, which landed in `cc14794` — `d1c4edd62` covers the same ground, so GitHub already has this branch as `CONFLICTING`.

Your diagnosis was right and matched mine independently: same root-cause pattern as #87, inherited Bitcoin mainnet WIFs failing to decode under BTQ's `SECRET_KEY` version, leaving an invalid `CKey` whose `GetPubKey()` trips `assert(keydata)`. Worth adding one detail you may not have seen from the `key_test1` abort alone: the aborted fixture left the logger mid-initialisation, so the *next* case died on `assert(m_buffering)` and the binary then hung rather than exiting — which is why the failure looked like a hang instead of an abort depending on where you stopped.

Verified on master's tree just now, all four cases run:

```
key_tests.cpp(35):  Entering test case "key_test1"
key_tests.cpp(157): Entering test case "key_signature_tests"
key_tests.cpp(203): Entering test case "key_key_negation"
key_tests.cpp(347): Entering test case "key_ellswift"
--- test_btq exit code: 0
```

Thanks for jumping on it — your note on #91 that `key_tests` was the next fatal in the chain is what pointed me at it in the same pass.

augchan42:
Superseded by #91 — agreed, and worth recording *why* the overlap is reassuring rather than just wasteful.

The eight replacement literals in #91 are **byte-for-byte identical** to the ones here, including `strAddressBad`, which was the one value not forced by the crash. Leaving that as the Bitcoin address would still have passed, but only accidentally — it would test that a *foreign-chain* string is not a private key, where the upstream intent was that a well-formed address of *the same chain* is not one. Two derivations arriving at the same answer independently, including on that judgement call, is a stronger check on those values than either of us reviewing the other's.

One thing that did not carry over, offered rather than asserted: this branch put the derivation method in the file itself — that only the base58 version byte moves, that the 32 secret bytes and compression flag are unchanged so every downstream pubkey/signature/ellswift expectation still holds, and that the values came from `test/functional/test_framework` with the pipeline validated by reproducing all eight *original Bitcoin* literals first. #91 has the correct values without that note.

Worth porting the comment if you agree, purely so the next person can tell these are re-derivable rather than magic — same reasoning as the generator note you put in the BIP324 commit. Happy to open a docs-only PR against master after #91 lands if useful.

Verified on #91's head (`d1c4edd62`) locally: `key_tests` passes, 6 cases, 0 failures.

Issue #95, opened 2026-07-30
p2p: v2 transport is disabled by default and the intended transport end state is unstated


BarneyChambers:
## Summary

Two related observations about the P2P transport, both raised by @augchan42 in #92. That PR is closed (its BIP324 vectors were regenerated in #91 rather than skipped), but these two points outlive it and I didn't want them to disappear with the branch.

## 1. v2 transport is off by default here, and on by default upstream

`src/net.h:111`:

```cpp
static constexpr bool DEFAULT_V2_TRANSPORT{false};
```

Upstream Bitcoin Core ships this as `true`. It feeds `-v2transport` (`init.cpp:502`) and gates whether the node advertises `NODE_P2P_V2` (`init.cpp:899`), so as shipped, BTQ nodes neither offer nor advertise the encrypted transport, and `src/bip324.cpp` is inert in the default configuration.

I could not find this recorded as a decision anywhere. Two things worth separating:

- If it is deliberate, the reasoning should be written down next to the constant. It is the kind of thing an auditor will flag as a downgrade from upstream and it is cheaper to answer once, in the code, than in an audit finding.
- If it is an accident of the fork (for instance the default was simply never revisited after the BIP324 vectors started failing), it should be flipped, since the transport itself is intact — #91 established that by regenerating the vectors against an independent reference implementation and getting a passing suite.

## 2. BIP324's handshake is not post-quantum, and the fork should say what it intends to do about that

BIP324's key agreement is ElligatorSwift over secp256k1 ECDH. For a chain whose whole premise is post-quantum signatures, an encrypted transport whose forward secrecy rests on the discrete log problem is an odd foundation, and it is a fair question whether effort belongs in reaching parity with upstream BIP324 or in a transport designed against the same threat model as the rest of BTQ.

To be explicit about what this is *not*: "not post-quantum" is not on its own an argument for removing or neglecting BIP324. It carries real value against classical adversaries — transaction-origin linkage and DPI-based censorship — and by the standard of "drop anything not post-quantum" the ECDSA compatibility paths would go too. @augchan42 made this point in #92 and I agree with it.

The question is narrower: **what is the intended end state?** Reasonable answers include keeping BIP324 as-is for classical privacy and accepting that a store-now-decrypt-later adversary can retrospectively read traffic; or planning a PQ or hybrid handshake. Either is defensible. Having no stated position is the thing that will cost time in the audit, because an auditor has to guess at the threat model before they can judge the implementation.

[...]

Issue #94, opened 2026-07-30
consensus/policy: script and block resource limits were raised for Dilithium, but the intended envelope is not written down


BarneyChambers:
## Summary

BTQ raises several script and block limits to accommodate Dilithium, but some closely-coupled limits were left at Bitcoin's values. The combination changes the interpreter's worst-case resource envelope, and I don't think the intended envelope is written down anywhere. With an external audit being scheduled, it would be better to settle this deliberately than to have it inferred from a set of raised constants.

This came out of @augchan42's review on #91, where the immediate finding was that the P2SH size guard in `descriptor.cpp` is now unreachable and its comment was three orders of magnitude out of date. That comment is corrected separately; this issue is the broader question underneath it.

## What changed and what didn't

Consensus:

| Constant | Bitcoin | BTQ | File |
| --- | --- | --- | --- |
| `MAX_SCRIPT_ELEMENT_SIZE` | 520 | **15000** | `script/script.h:29` |
| `MAX_SCRIPT_SIZE` | 10000 | **100000** | `script/script.h:43` |
| `WITNESS_SCALE_FACTOR` | 4 | **16** | `consensus/consensus.h:21` |
| `MAX_BLOCK_WEIGHT` | 4000000 | **8000000** | `consensus/consensus.h:15` |
| `MAX_BLOCK_SERIALIZED_SIZE` | 4000000 | **8000000** | `consensus/consensus.h:13` |
| `MAX_OPS_PER_SCRIPT` | 201 | 201 (unchanged) | `script/script.h:32` |
| `MAX_STACK_SIZE` | 1000 | 1000 (unchanged) | `script/script.h:46` |
| `MAX_BLOCK_SIGOPS_COST` | 80000 | 80000 (unchanged) | `consensus/consensus.h:17` |

Standardness:

| Constant | Bitcoin | BTQ | File |
| --- | --- | --- | --- |
| `MAX_STANDARD_P2WSH_STACK_ITEM_SIZE` | 80 | **15000** | `policy/policy.h:46` |
| `MAX_STANDARD_TAPSCRIPT_STACK_ITEM_SIZE` | 80 | **15000** | `policy/policy.h:48` |
| `MAX_STANDARD_SCRIPTSIG_SIZE` | 1650 | **15000** | `policy/policy.h:52` |
| `MAX_STANDARD_P2WSH_SCRIPT_SIZE` | 3600 | **65536** | `policy/policy.h:50` |
[...]

PR #96, opened 2026-07-30
test: get the unit binary to run to completion

BarneyChambers:
## Summary

Follow-on to #91. That PR stopped `descriptor_tests` segfaulting and got the run from 44 suites to 105; this one clears the inherited-data failures it exposed and fixes the reason the binary still never finished.

**The headline: the unit test binary now runs to completion.** It previously had to be killed.

| | at `cc14794` (master) | here |
| --- | --- | --- |
| Whole binary, one process | hangs, killed | **completes, 38s** |
| Failing suites (run in isolation) | 15 | **9** |
| Suites that hang or abort | 3 | **0** |

The 9 that remain are red on purpose and are listed below. None of them are things I could honestly call stale data, and I would rather leave them visible than paper over them before an audit.

## What changed

**Inherited upstream data (6 suites).** Same root cause as #91 in most cases.

- `util_tests` — BTQ sets `MESSAGE_MAGIC` to `"BTQ Signed Message:\n"` (`util/message.cpp:24`), so every signed-message vector moves, and mainnet base58 prefixes put all four addresses out of range. Regenerated from RFC6979 ECDSA independently of this implementation: the generator was first run with upstream's magic and versions and required to reproduce all five upstream values byte for byte. It then agreed with the C++ output independently, which is a check from both directions. Also `test_Capitalize`, which is not inherited data but a find-replace accident — `Capitalize("bitcoin") == "Bitcoin"` had become `Capitalize("btq") == "BTQ"`, asserting that a function which upper-cases the first character upper-cases the whole string.
- `script_standard_tests` — 8 taproot addresses under the `qbtc` HRP. Fixed by swapping the HRP at test time rather than editing the addresses, so `test/data/bip341_wallet_vectors.json` stays byte-identical to the file published with the BIP and remains diffable against it. The witness program is unaffected and is still asserted directly against the unmodified vector.
- `rpc_tests`, `psbt_wallet_tests` — stale P2SH address, WIFs and xprv, re-encoded with payloads carried across untouched and each conversion round-tripped back under Bitcoin's version bytes.
- `walletload_tests` — worth a second look. The descriptor written to provoke an ID mismatch carried a Bitcoin xpub, so `LoadWallet()` returned `UNKNOWN_DESCRIPTOR` before it ever compared IDs. The case was passing through a path it was not written to test.
- `script_parse_tests` — not stale data. BTQ added five Dilithium opcodes at `0xbb`–`0xbf` and moved `MAX_OPCODE` from `OP_NOP10` to `OP_DILITHIUM_PUBKEY`, which brings `OP_CHECKSIGADD` (`0xba`) back into the parser's table, so upstream's assertion that it is unknown cannot hold. The unknown-opcode path is still covered with a name that cannot become valid, and the widening is now asserted explicitly — the same constant also bounds `CScript::HasValidOps()`.

**The hang.** `coinselector_tests` exceeded a 180s timeout and had to be killed. No individual case is slow; every one finishes in under two seconds. It was a cascade: `srd_tests` fails, `check_max_weight` then calls `result->GetInputSet()` on an empty `Result`, `util::Result::value()` asserts, and the abort happens inside a wallet fixture. The fixture never tears down, so `CCheckQueue`'s destructor trips its own `assert(m_worker_threads.empty())` and the process wedges instead of exiting. `spend_tests/SubtractFee` reached the same state through `*res`. Three dereferences promoted to `BOOST_REQUIRE`.

This fixes none of the underlying failures — they are all still red — but they are now *reported*, and two of them answer themselves once they stop crashing:

```
spend_tests/SubtractFee: CreateTransaction failed: Insufficient funds
descriptor_p2sh_segwit_tests: dilithium-bech32 is disabled because witness v0
keyhash programs are indistinguishable from ECDSA P2WPKH and are not spendable
with Dilithium keys
```

That second one is not a defect. It is a deliberate decision, refused in all three places that could produce such an address, and the test was contradicting it — handing out that address would create unspendable funds. Rewritten to assert the refusal so the decision is covered rather than fought. What remains in that case is real: `CalculateMaximumSignedInputSize()` returns nothing usable for a `dilithium-legacy` output, so fee estimation for it does not work.

**One production file, comment only.** `descriptor.cpp` claimed its P2SH size check "limits the maximum number of compressed pubkeys to 15" — true at Bitcoin's 520-byte `MAX_SCRIPT_ELEMENT_SIZE`, three orders of magnitude out at BTQ's 15000. Raised by @augchan42 on #91. The guard is unreachable: `multi()` caps at 20 keys and the only `PubkeyProvider` implementations return 33 or 65 bytes, since Dilithium reaches descriptors through `AddressDescriptor` rather than `multi()`, so `script_size` cannot exceed 1320. The check is left in place; only the comment changes.

## What is still red

All nine were red before this PR and none are caused by it.

| Suite | Failures | Character |
| --- | --- | --- |
| `miner_tests/CreateNewBlock_validity` | 112 | `bad-txns-vout-negative` — a negative output value |
| `txpackage_tests` | 21 | package limits and `bad-txns-in-belowout` |
| `validation_chainstatemanager_tests` | 10 | `CreateAndActivateUTXOSnapshot` fails |
| `miniscript_tests/fixed_tests` | 5 | scripts upstream rejects now parse |
| `validation_tests/test_assumeutxo` | 4 | expected hashes vs what looks like garbage |
| `coinselector_tests` | 3 | coin selection returns nothing |
| `validation_chainstate_tests` | 1 | same snapshot path |
| `spend_tests/SubtractFee` | 1 | `Insufficient funds` |
| `descriptor_p2sh_segwit_tests` | 1 | Dilithium fee estimation |

Two I would flag as deserving attention rather than triage: `miner_tests` failing on a **negative transaction output value**, and `validation_tests/test_assumeutxo` where the observed hash (`0000001c000055e40d082850…`, and `32764` for a chain tx count) does not look like a wrong-but-deterministic value so much as an uninitialised read. Both are being investigated separately and neither is in scope here.

## Related

- #92 and #93 closed as superseded — #92 would have applied a skip to `bip324_tests`, which #91 made pass.
- #94 filed on the script and block resource limits, which is the broader question underneath the `descriptor.cpp` comment.
- #95 filed on `DEFAULT_V2_TRANSPORT{false}` and the intended transport end state, preserving the points from #92.

## Test plan

- Whole binary runs to completion in one process: 38s, exit 201, 158 failures, no hang and no abort
- The 6 suites fixed here report zero failures
- `bip324_tests`, `descriptor_tests`, `key_tests`, `key_io_tests`, `denialofservice_tests` from #91 still pass
- Signed-message generator reproduces all 5 upstream values under upstream's magic before emitting BTQ's
- Every base58 re-encoding round-trips back to the original under Bitcoin's version bytes
- Descriptor checksum tool reproduces the existing literal before generating a new one
- One production file changed, comment only


BarneyChambers:
The unit suite is now green. `test_btq` runs 733 test cases to completion with no errors, from 159 failures when this branch opened.

Four more commits, one per suite:

**coinselector_tests (3).** Three cases carried weight arithmetic derived from Bitcoin's `WITNESS_SCALE_FACTOR` of 4. A P2WPKH input costs 768 weight units here rather than 272, because an input's base bytes are multiplied by the scale factor and ours is 16, so a standard transaction holds about 519 inputs rather than about 1470. Two cases wanted selections that no longer fit and one relied on a fee that shrank. Rewritten against the real per-input weight, and checked over 15 consecutive runs since the solver is randomised. One of them, `srd_tests` case 1, had been passing for the wrong reason: its budget no longer covered the coins it was meant to accommodate, so it was failing on weight rather than on the insufficient funds it tests.

**spend_tests (1).** `SubtractFee` paid out a literal 50 BTQ, so every case asked for ten times what the fixture's coinbase held. It now reads the coinbase's value from the wallet rather than restating a subsidy that can move again. `wallet_duplicated_preset_inputs_test` was passing but degenerately, with a target so far above the balance that the double-counting it guards against could not have closed the gap; its amounts are back in proportion.

**miniscript_tests (5).** All five asserted a script was invalid under P2WSH, and all five got that verdict from one limit rather than the three the comments describe: `TESTMODE_P2WSH_INVALID` drives `IsValid()`, which checks only type and script size. The five scripts run from 3759 bytes to about 5 kB, over Bitcoin's 3600-byte ceiling for a witnessScript and under ours of 65536. Dropping the flag runs them through the valid branch in both contexts, which covers more than the assertion it replaces, since the ops and stack ceilings are unchanged and `TestSatisfy` pins down exactly how the P2WSH satisfactions must fail. A new case covers the size ceiling at the value we actually set.

**txpackage_tests (21).** This file was converted halfway: `parent3` already spent a 5 BTQ coinbase while everything around it still paid out from a 50 BTQ subsidy, so most of the suite failed on `bad-txns-in-belowout`. Two limits had also moved out from under their cases. `MAX_PACKAGE_WEIGHT` is 40,000,000 WU rather than 404,000, so the package built to exceed it tripped the count limit first and was rejected as `package-too-many-transactions`. And the giant-transaction premise asserted a size above `DEFAULT_ANCESTOR_SIZE_LIMIT_KVB`, which upstream can do because that limit and `MAX_STANDARD_TX_WEIGHT` both sit near 100 kB; here they are 2 MB and 25,000 vbytes, so the premise now names the limit that actually produces the `tx-size` rejection it sets up.

### One real defect, not stale data

Everything above is inherited test data. One failure was not, and is filed as #97: the wallet hands out `dilithium-legacy` destinations on chains where they are neither valid nor spendable. `nDilithiumP2MRHeight` is 1 on mainnet, signet and regtest, so `LegacyDilithiumBase58PaymentsAllowed()` is false and `IsValidDestination` rejects the destination, yet `GetNewDestination` still returns one. No descriptor models a Dilithium P2PKH and the dummy-signing fallback only sizes witness satisfactions, so `CalculateMaximumSignedInputSize` returns -1 and `AvailableCoins` marks any such output unsolvable. This is the hazard `dilithium-bech32` was explicitly disabled for; the legacy variant has no equivalent guard.

Rather than delete the case, `mixed_wallet_input_fee_estimation` now asserts both halves of the present behaviour, so the defect is visible from the suite and not only from the tracker, and fixing either half fails there and prompts the case to be restored to the positive assertion it wants to make.

### Also filed

A capacity note on #94. Raising `WITNESS_SCALE_FACTOR` to 16 and `MAX_BLOCK_WEIGHT` to 8,000,000 while leaving `MAX_STANDARD_TX_WEIGHT` at 400,000 moves capacity in opposite directions: witness bytes per block double to 8,000,000, non-witness bytes per block halve to 500,000, and non-witness bytes per standard transaction drop to a quarter of upstream's at 25,000. That shape suits a chain carrying ~3.3 kB signatures in the witness, but the three constants were changed independently and the interaction between them is not documented anywhere in the tree.

BarneyChambers:
Three more commits, all of them about the test suite being able to run rather than about what it asserts.

**`make check` was stopping partway through, on master and on every branch.**

The per-file rule in `src/Makefile.test.include` extracts the suite names from each source file and hands them to Boost's `-t`, joined with newlines. Boost takes one comma-separated list there, so a file declaring two suites produced a filter matching neither:

```
Running tests: spend_tests descriptor_p2sh_segwit_tests from wallet/test/spend_tests.cpp
Test setup error: no test cases matching filter or all test cases were disabled
```

`wallet/test/spend_tests.cpp` is the only file in the tree declaring two suites, which is why one missing separator was enough. Upstream never hit it: `descriptor_p2sh_segwit_tests` is ours, added alongside the inherited `spend_tests`.

The consequence was larger than a red gate. `make check` stopped there, so everything ordered after it — the rest of `wallet/test`, and the `btq-tx` util runner — has never executed under CI at all.

**The `btq-tx` util tests had never run either.**

`test/util/data/btq-util-test.json` named `./bitcoin-tx` and `./bitcoin-util` as the program to run in all 95 cases. Those paths do not exist here, so every case failed to start. With the names corrected, 39 still failed, for three reasons:

- Seven cases pass a Bitcoin mainnet WIF and nine pass mainnet addresses, none of which decode under our version bytes. `btq-tx` rejects them up front, so five cases that exist to check a specific error message were reporting `privatekey not valid` instead of the message under test. Re-encoded with the payload carried across untouched.
- Every expected `weight` was a quarter of what it should be, `WITNESS_SCALE_FACTOR` being 16 here rather than 4. Recomputed from the serialized transaction, and each old value confirmed to be the same figure under Bitcoin's factor.
- The decode output renders `scriptPubKey.address` and `.desc`, which move with the prefix even though the script does not.

That the eight `.hex` expectations then matched byte for byte is the check on the first step: same key, same scripts, same signatures, same serialization. 95 of 95 now pass.

**The GUI suite had one failure.**

`rpcNestedTests` compared the tip's coinbase txid against a hardcoded literal that is Bitcoin's genesis coinbase — which our genesis cannot produce, carrying a different message and subsidy. Read from the chain parameters instead. `test_btq-qt` is now 21 passed, 0 failed.

---

**CI**

`build-and-address-tests` passes. It now runs `make check` to completion for the first time.

`test each commit` fails and cannot pass from this branch. It rebuilds and fully tests each of the last six commits independently, so a branch that fixes tests in stages is red in the middle by construction — it failed on `1ff7bcea0` at the Qt test, whose fix lands three commits later. It also runs the whole functional suite on every commit, and that suite is not green: it carries the same inherited constants this PR has been working through, e.g. `rpc_blockchain.py` asserts a UTXO total of 8725 BTC, which is Bitcoin's 50-per-block subsidy against our 5. That is a separate migration, not addressed here.

Worth noting for anyone reading the checks: `test each commit` only runs on pull requests with more than one commit and never on master, and the macOS and Win64 jobs are disabled. So the lean gate is what actually guards master, and it is green.

PR #98, opened 2026-07-30
test: fix miner_tests for BTQ activation heights, LWMA and subsidy

augchan42:
## Summary

`miner_tests` is the first failing file the CI merge gate reaches on master, so nothing sorted after it runs there. It reported 112 failures. **It now passes.**

Tracked as BTQ-112 under BTQ-90.

## These were not 112 problems

They were four places where this file encodes **upstream Bitcoin's consensus activation schedule**, plus a subsidy mismatch. Every one has the same shape: *BTQ activates at height 1 what Bitcoin activates at a fork height*, so code paths that lie dormant across these test heights upstream are live here.

Each was read off validation's own rejection message rather than inferred.

| # | Rejection | Blocks | Cause |
|---|---|---|---|
| 1 | `unexpected-witness` | 45 | `SegwitHeight = 1` vs Bitcoin's 481824 |
| 2 | `bad-diffbits` | 65 | `nLWMAHeight = 1`, retargets per block vs Bitcoin's per-2016 |
| 3 | difficulty ramp | — | `MedianTimePast+1` pacing under per-block LWMA |
| 4 | `bad-txns-nonfinal` | — | `CSVHeight = 1` vs Bitcoin's 419328 |

**1 — unexpected-witness.** With segwit live at these heights, `CreateNewBlock` adds a witness commitment output *and* the coinbase witness reserved value. The inherited `vout.resize(1)` drops the commitment but left the witness, so the block carried witness data while committing to none. Now clears the coinbase witness alongside it.

**2 — bad-diffbits.** The loop reuses one template for all 110 blocks. Bitcoin retargets once per 2016 blocks, so a single `nBits` is valid throughout; BTQ's per-block LWMA makes the height-1 value go stale as the chain advances. Now recomputed per block.

**3 — difficulty ramp.** Pacing at `MedianTimePast+1` reads, under per-block LWMA, as the chain running ~60x fast. Difficulty climbed until the nonce search stalled — reproducibly around block 45. Now paced at `nPowTargetSpacing`, which holds LWMA at steady state and keeps these blocks cheap to mine. **If you change that pacing, the nonce table must be regenerated** — noted in-file.

**4 — bad-txns-nonfinal.** Upstream holds relative-locked transactions in the mempool and still assembles a valid template; its comment says so outright — *"these will still generate a valid template until BIP68 soft fork"*. With `CSVHeight = 1` that template is invalid here, and the assembler does not filter by sequence locks because `addUnchecked` bypassed the mempool checks that normally would. Those two transactions are now added once the chain has aged enough for them to be final, and the intermediate assertion reflects what BTQ actually produces.

## BLOCKINFO regenerated

BTQ has its own genesis, so no inherited nonce satisfies any block hash in this chain. The regeneration procedure is recorded above the table since it must be re-run if block contents change.

Worth flagging for reviewers: **the nonces could not be regenerated before 1–3 were fixed.** While blocks were being rejected the chain never advanced, so every coinbase carried `scriptSig` height 1 and the emitted table was meaningless. I generated exactly such a table on the first attempt and nearly committed it. If you regenerate, confirm the run reports zero `ProcessNewBlock` failures first.

## Sigop cases derived, not hardcoded

Upstream's 1001 transactions was calibrated for Bitcoin's limit and is wrong here twice over:

- it no longer sits at the sigop boundary
- `1001 × LOWFEE` overdraws BTQ's 5 BTQ subsidy, driving the chained outputs negative from iteration 500 and failing `bad-txns-vout-negative` long before any sigop limit is reached

The count now derives from `MAX_BLOCK_SIGOPS_COST / WITNESS_SCALE_FACTOR` (251 at present), which keeps the fees inside the subsidy as a consequence rather than by tuning. The declared cost moves from `SigOpsCost(80)` to `320` — the literal 80 is Bitcoin's `20 × 4`, not BTQ's `20 × 16` — so the assembler's budget again agrees with what validation recomputes.

Deriving rather than hardcoding also means this test does not depend on BTQ-118 being resolved: if `MAX_BLOCK_SIGOPS_COST` or `WITNESS_SCALE_FACTOR` change, it follows.

The remaining `5000000000` literals are Bitcoin's 50 BTC subsidy against BTQ's 5 and spent ten times what the inputs hold; now a named constant.

## Result

- `miner_tests`: **112 failures → 0**
- Full binary at CI parity: **171 → 59 failures**, no regressions elsewhere

Still terminates at `psbt_wallet_tests` — an inherited Bitcoin `xprv` (`0488ade4` vs BTQ's `0488ade5`), separate and next.

## Test plan

- [x] `test_btq -t miner_tests` → `*** No errors detected`, exit 0 (was exit 201, 112 failures)
- [x] Full binary re-run: 119 suites / 665 cases, 59 failures, none in `miner_tests`
- [x] Each root cause confirmed from validation's rejection message, not inferred
- [x] Regenerated nonces produced from a run with zero `ProcessNewBlock` failures
- [ ] CI will still be red — expected; it should now advance past `miner_tests` to the next failing file

## Note for reviewers

Given four activation-height divergences in one file, I would expect the same class in the remaining suites. Worth checking `SegwitHeight` / `CSVHeight` / `BIP65Height` / `BIP66Height` assumptions before concluding that any inherited test needs its vectors regenerated.

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -8,8 +8,10 @@
 #include <consensus/consensus.h>
 #include <consensus/merkle.h>
 #include <consensus/tx_verify.h>
+#include <chainparams.h>
 #include <node/miner.h>
 #include <policy/policy.h>
+#include <pow.h>
 #include <test/util/random.h>
 #include <test/util/txmempool.h>
 #include <timedata.h>
@@ -58,6 +60,11 @@ BOOST_FIXTURE_TEST_SUITE(miner_tests, MinerTestingSetup)
 
 static CFeeRate blockMinFeeRate = CFeeRate(DEFAULT_BLOCK_MIN_TX_FEE);
 
+// Value of the txFirst coinbases these tests spend. BTQ pays 5 BTQ per block
+// where upstream Bitcoin pays 50, so the inherited 5000000000 literals spend ten
+// times what the input holds and fail with bad-txns-in-belowout.
+static constexpr CAmount COINBASE_VALUE = 5 * COIN;
+
 BlockAssembler MinerTestingSetup::AssemblerForTest(CTxMemPool& tx_mempool)
 {
     BlockAssembler::Options options;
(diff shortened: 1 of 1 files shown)


augchan42:
Closing as superseded by #96, which covers this file and the rest of the suite.

Verified #96 locally at CI parity (`d1c4edd62` base, branch head): **133 suites, 733 cases, 0 failures, exit 0** — the unit binary runs to completion and is green. This branch got it to 59 failures and still aborted at `psbt_wallet_tests`, so #96 is strictly further along.

We reached the same four divergences independently and within minutes of each other, which is worth recording since it is the third time today. For anyone reading later, the agreement is the useful part: `SegwitHeight = 1` vs 481824, `nLWMAHeight = 1` retargeting per block vs per 2016, the LWMA ramp from `MedianTimePast+1` pacing, and `CSVHeight = 1` vs 419328 making BIP68 live where upstream's own comment assumes it dormant.

Two things #96 does better, and I would not want them lost by treating this as a coin flip:

**The BIP68 assertion.** I deferred the relative-locked transactions so the template came out valid and small. #96 asserts the template *is rejected*, which exercises BTQ's enforcement instead of routing around it. That is a stronger test and it covers more than upstream can.

**The fee scaling.** #96 keeps upstream's 1001-link chains and scales every fee to a tenth, matching the tenth-size subsidy, so the proportions the test depends on are preserved. I changed the link count instead, which perturbs the structure more than the subsidy change required.

One correction from my side that is already fixed in #96's `descriptor.cpp` commit: I wrote on BTQ-118 that a 20-key `multi()` redeemScript tops out around 700 bytes. That assumed compressed keys only. Accounting for 65-byte uncompressed providers it is 1320, as #96 states. Same conclusion — nowhere near `MAX_SCRIPT_ELEMENT_SIZE` — but my figure was wrong.

Nothing here needs porting. The one thing I would carry forward is a process point rather than code: this is the third PR of mine today superseded within minutes by concurrent work on the same file (#92, #93, #98). Worth a word on who picks up what before starting, since the board does not signal it.

PR #99, opened 2026-07-30
Add Dilithium PSBT (BTQ-PSBT) design specification

Oskii:
## Summary

Adds `design/DILITHIUM_PSBT_DESIGN.md` — a draft v3 design spec for **BTQ-PSBT**, extending BIP174 PSBT to support Dilithium partial signatures on P2MR (witness v2) outputs.

This is a **design-only PR** for community review before implementation. No consensus or code changes.

The spec covers:

- New PSBT input/output type bytes (`0x19`–`0x1D`, `PSBT_OUT_P2MR_TREE`)
- Normative sighash (BIP341 `SignatureHashSchnorr` + `P2MR_TAPSCRIPT`)
- Normative `VerifyP2MRCommitment` and accumulator multisig witness assembly (production `btq-multisig` path)
- Merge semantics (`combinepsbt` / `joinpsbts`), fail-closed validation, and error taxonomy
- Migration path from interim `.btqms` JSON format
- Required positive/negative test vectors (Appendix B) for Phase 0 implementation gate

## Why now

Today, Dilithium P2MR spends work on-chain and via `btq-multisig` `.btqms`, but btq-core PSBT drops Dilithium fields at serialization. This spec defines the interchange format so wallets, co-signers, and external tools can interoperate without a hardfork.

## Open for discussion

- Type byte allocation and field naming
- v1 sighash scope (ALL only; NONE/SINGLE/ANYONECANPAY deferred)
- Single `leaf_hash` per input constraint
- Proprietary `"btq.org"` fields marked out of v1 production scope
- Whether `design/` is the right home vs `doc/design/`

## Test plan

- [ ] Protocol/wallet reviewers read spec and comment on wire format (§4) and finalization (§6.5)
- [ ] Confirm accumulator witness stack matches `btq-multisig/finalize.cpp` expectations
- [ ] No code changes in this PR — implementation follows in a separate PR after spec approval


augchan42:
Reviewed against source at `5ce8e10bd` rather than reading it through. This is a strong spec — the normative sections are unusually complete for a v1 design doc, and §2.5 is precise enough to implement from without reading btq-core, which is what it claims.

I have a narrow but relevant vantage: I built a working P2MR spend against the interpreter earlier today (`bench/dilithium_verify.cpp`, #100), so §2.5 is checkable against something that actually verifies.

**§2.5.2 and §2.5.3 are correct**, and confirmed the hard way. My first attempt aborted with `Missing data` until I passed `force=true` to `PrecomputedTransactionData::Init` — exactly the requirement §2.5.2 calls out and explains. The `execdata` field set in §2.5.3 matches what I had to set for `VerifyScript` to return `SCRIPT_ERR_OK`. Good.

Three findings below. The first two are concrete errors; the third is a design question.

---

## 1. `MAX_CONTROL_BLOCK_SIZE` is wrong by 8x (§4.5)

The table gives:

| `MAX_CONTROL_BLOCK_SIZE` | 513 | `P2MR_CONTROL_MAX_SIZE` |

It cites the source constant by name, but that constant is not 513:

```cpp
// script/interpreter.h:252-255
static constexpr size_t P2MR_CONTROL_BASE_SIZE  = 1;
static constexpr size_t P2MR_CONTROL_NODE_SIZE  = 32;
static constexpr size_t P2MR_CONTROL_MAX_NODE_COUNT = 128;
static constexpr size_t P2MR_CONTROL_MAX_SIZE   = 1 + 32 * 128;   // = 4097
```

513 is `1 + 32*16`, i.e. a 16-node tree. Consensus permits **128**.

A parser enforcing 513 fail-closed, as §4.5 instructs, would reject control blocks for any tree deeper than 16 levels — spends that are valid on-chain. Suggest 4097, or state explicitly that v1 imposes a tighter policy cap than consensus and why.

## 2. `MAX_P2MR_LEAF_SCRIPT_SIZE` cannot hold the spec's own headline use case (§4.5)

The cap is 10,000 bytes. Consensus allows `MAX_SCRIPT_SIZE = 100,000` (`script/script.h:43`).

That gap matters more here than it would on Bitcoin, because a Dilithium pubkey is 1,312 bytes. Rough leaf sizes for `n`-of-`n`:

| leaf | approx size | under 10,000? |
|---|---|---|
| single key | 1,314 | yes |
| 5-of-5 | 6,569 | yes |
| 7-of-7 | 9,195 | yes |
| **8-of-8** | **10,508** | **no** |
| 20-of-20 | 26,264 | no |

So the format cannot represent an 8-key leaf, while consensus would accept roughly 76. Given §2.2 is titled *"Production multisig: accumulator leaves (critical)"*, a limit that stops at seven keys seems likely to bind on exactly the case the spec exists to serve.

If 10,000 is deliberate — bounding parser memory, say — it is worth saying so and stating the resulting key ceiling explicitly, because implementers will otherwise discover it at 8-of-8.

## 3. §2.5.5 contradicts §4.6 on the finalized sighash byte

§4.6 is right, and clear: *"Finalizer MUST emit `0x01` in witness."* That is necessary, because the interpreter rejects `SIGHASH_DEFAULT` outright for this sigversion:

```cpp
// script/interpreter.cpp
if (sigversion == SigVersion::P2MR_TAPSCRIPT) {
    if (nHashType == SIGHASH_DEFAULT) return false;
```

Worth calling out prominently, incidentally: this is a **deliberate divergence from BIP341**, where `SIGHASH_DEFAULT` is the normal case and is implied by an absent byte. On P2MR the byte is mandatory and `0x00` is invalid.

But §2.5.5 defines, under a MUST heading:

> `sig_with_hashtype` | 2421 bytes | `sig_raw || sighash_type_byte`. Stored as PSBT value; **appended to witness stack on finalize**.

with no reference to normalization. An implementer working from §2.5.5 — which is where the witness construction is specified — appends the wire byte. If that byte is `0x00`, the result is a transaction that fails validation with no diagnostic pointing at the sighash byte.

Suggest §2.5.5 state the finalized byte is the normalized one, or simply defer to §4.6.

**Underlying design question:** why accept `0x00` at all? v1 supports only `SIGHASH_ALL`, the interpreter rejects `0x00`, and accepting it buys a normalization step, a cross-section consistency rule, and a failure mode. Rejecting `0x00` at parse — matching the interpreter — removes all three. If the motivation is tolerating signers that default to taproot conventions, that is a good reason and worth recording; otherwise the simpler rule looks strictly better.

---

## Smaller notes

- `MAX_DILITHIUM_PARTIAL_SIGS_PER_INPUT = 20` lines up with `MAX_PUBKEYS_PER_MULTISIG = 20`. Worth stating that is where it comes from, so the two move together.
- §2.5.4's transcript table (`ext_flag 1`, `key_version 0`, `spend_type 2`) matches BIP342 and the P2MR path. No issue.
- The 2420/2421 split in §2.5.5 matches `DilithiumConstants::SIGNATURE_SIZE`. Note both are Dilithium2-specific: `dilithium_key.h` also carries Dilithium5 constants (2592-byte key, 4627-byte signature), so if the parameter set ever moves, §4.5 and §2.5.5 are the places that silently become wrong. A sentence pinning v1 to Dilithium2 would help.

Happy to see this land once 1 and 2 are settled — they are the two an implementer would hit as runtime rejections rather than as spec ambiguity. Nothing here changes the architecture, which looks right, particularly the choice to mirror the taproot field layout rather than invent one.

augchan42:
Second pass, this time through an adversarial three-lane review (zero-context executability / grounding / structure-and-risk) run in fresh context by a different model, with every source claim re-verified by me before posting. It found things I missed, and it corrected one of my own numbers.

**Three critical issues. All three verified against the interpreter.**

---

## C1 — The `CHECKMULTISIGDILITHIUM` witness in §6.5.1 is unspendable for any `m<n`

§6.5.1 specifies one signature per pubkey:

```
stack[1] = sig_with_hashtype for pk1
...
stack[n] = sig_with_hashtype for pkn
```

But `OP_CHECKMULTISIGDILITHIUM` follows CHECKMULTISIG stack discipline: it reads `nSigsCount` from the script's `<m>` push and consumes exactly `m` signatures plus the dummy (`src/script/interpreter.cpp:1347-1393`). Supplying `n` leaves `n-m` elements on the stack, which fails P2MR clean-stack enforcement.

So a 2-of-3 built to this spec does not spend. Since `m == n` is the only case that works, this affects essentially every real multisig policy — and it contradicts §5.6, which declares an input finalizable with `m` valid signatures. The parenthetical in §6.5.1 ("CHECKMULTISIG requires `m` valid sigs on finalize, not empty slots") suggests the intent was already `m`; the stack diagram just does not match it.

Worth contrasting with `THRESHOLD_ACCUMULATOR` in §6.5, which correctly uses fixed slots with empty vectors — the two templates have genuinely different stack shapes and the document is right to separate them.

**Suggested fix:** specify `[empty_dummy, sig_1..sig_m, leaf_script, control_block]` where the `m` signatures are an ordered subsequence matched to increasing pubkey positions, restate the CHECKMULTISIG matching rule, and add vectors for non-consecutive signers (1-of-3 signing with pk3; 2-of-3 signing with pk1+pk3).

## C2 — `FinalizeP2MRInput` accepts an unverified pre-existing witness

```
FinalizeP2MRInput(input, tx, index, utxo):
  if final_script_witness already set: return OK
  ...
  VerifyScript(..., STANDARD_SCRIPT_VERIFY_FLAGS) MUST succeed
```

The early return sits above the verification. An imported PSBT carrying a malformed or cryptographically invalid `final_script_witness` is therefore reported finalized and will extract to an unspendable transaction. §6.3's classifier has the same shape — presence implies `FINALIZED`, even for an unknown leaf.

This is the one place the document's fail-closed posture inverts: everywhere else it verifies before trusting.

**Suggested fix:** verify any pre-existing final witness against the unsigned tx and resolved UTXO before returning `OK` or classifying as `FINALIZED`, and add negative fixtures for both malformed and cryptographically-invalid imported witnesses.

## C3 — §4.7 never ties `leaf_version` to the control byte

§4.7 step 3 computes the leaf hash from the *value's* `leaf_version`, and §4.9.2 does not compare it to the control byte either. Consensus derives it from the control byte:

```cpp
// src/script/interpreter.cpp:2268
execdata.m_tapleaf_hash = ComputeTapleafHash(control[0] & TAPROOT_LEAF_MASK, script);
```

A PSBT whose value-side `leaf_version` disagrees with `control_block[0] & 0xFE` therefore validates off-chain against one leaf hash and executes on-chain against another. A later list in the document does mention the comparison, so the normative sections disagree with each other.

**Suggested fix:** add `leaf_version == (control_block[0] & TAPROOT_LEAF_MASK)` as a mandatory step *before* hashing in §4.7, put it in `VerifyP2MRCommitment` itself, and add a mismatched-version negative vector.

---

## Correcting my own earlier comment

I previously wrote that the leaf-script ceiling is `MAX_SCRIPT_SIZE` (100,000), giving ~76 keys. **That was wrong.** The leaf script travels as a witness stack element, so the binding limit is `MAX_SCRIPT_ELEMENT_SIZE`:

```cpp
// src/script/interpreter.cpp:2075
if (elem.size() > MAX_SCRIPT_ELEMENT_SIZE) return set_error(serror, SCRIPT_ERR_PUSH_SIZE);
```

Corrected figures, at 1312 bytes per Dilithium pubkey:

| bound | value | max n-of-n |
|---|---|---|
| `MAX_P2MR_LEAF_SCRIPT_SIZE` (spec v1) | 10,000 | **7** |
| `MAX_SCRIPT_ELEMENT_SIZE` (consensus) | 15,000 | **11** |

The finding stands — v1 cannot express 8-of-8, which §2.2 calls the critical production case — but the gap is 4 keys, not 69. Apologies for the noise.

Conversely, my `MAX_CONTROL_BLOCK_SIZE` figure holds. The review's suggested 4,129 is taproot's (`33 + 32*128`, the 33 including an internal key P2MR has no equivalent of). P2MR is `1 + 32*128 = 4097` (`interpreter.h:251-255`), so the spec's 513 is still 8x low.

---

## Also worth your attention (not re-verified by me, judgement calls for you)

- **§4.3** — "depths strictly increase per BIP371 DFS rules" is not what `P2MRBuilder::ValidDepths` implements; siblings share a depth and DFS returns to shallower ones. The stated rule would reject valid trees.
- **§2.5.2** — requiring every spent output's amount `> 0` has no counterpart in source precomputation and makes a zero-value P2MR input unsignable. Use "known and valid `CAmount`", not zero-as-sentinel.
- **§5.5** — the leaf parser does not bound `m`/`n`, define ScriptNum limits, or reject duplicate pubkeys; since partial sigs are keyed by pubkey, a duplicated key could satisfy two accumulator slots.
- **§5.3.6 / §6.4** — multiple leaf entries are permitted but the signer algorithm is singular, and no selector exists. v1 should require exactly one leaf per input being signed, or define selection and merge.
- **§4.5** — parse-time verification is mandatory but unbounded PSBT-wide; the 1.5 MB per-input cap never binds (20 × 2421 = 48 KB). A large multi-input PSBT is a cheap way to force thousands of Dilithium verifications.
- **Appendix A** — `2 × (1344 + 2421)` is ~7.5 KB, not 7.5 MB; the base64 figure is ~10 KB, not 10 MB. Three orders of magnitude, and it undercuts the transport rationale that rests on it.

## What held up

Worth saying, given the length of the list: the grounded parts are genuinely well grounded. Dilithium2 sizes, `SigVersion::P2MR_TAPSCRIPT`, `ScriptExecutionData`, the sighash extension flags, `SIGHASH_DEFAULT` rejection, the control-block constants, and the internal `(DilithiumPKHash, leaf_hash)` signature keying all check out against source. The BIP360 parity-bit rule is covered in four places including a negative fixture — I went looking for that as a gap and did not find one.

The architecture is right. The field set does carry what a remote signer needs. C1 and C2 are the two that would produce broken transactions in production, and both are contained fixes rather than redesigns.

Full review with all 20 findings and my disposition: `reviews/plan-review-20260731-063830-418145.md`.

augchan42:
Note on the red CI, since this is a docs-only PR and the failure is confusing: **it is stale, not real.**

The run (`30553493300`) fired at 14:48 UTC on Jul 30. #96 merged to master at 15:17 UTC. The failing step is `make check` — the unit suite, which was still red at the time this branch was cut and went green with #96.

`git merge-base --is-ancestor 5ce8e10bd docs/dilithium-psbt-design` confirms the branch predates it.

A rebase onto `5ce8e10bd` should clear it. #100 had the identical symptom from the same 30-minute window and I have just rebased it. Nothing in this PR causes the failure.

Issue #97, opened 2026-07-30
wallet: dilithium-legacy destinations are still issued on P2MR-only chains, where they are neither valid nor spendable


BarneyChambers:
Found while getting the unit suite to pass (#96). `wallet/test/spend_tests.cpp` fails with `fee estimate failed for dilithium-legacy`, and the cause turns out to be broader than the test.

## What happens

`DescriptorScriptPubKeyMan::GetNewDestination(OutputType::DILITHIUM_LEGACY)` succeeds and returns a `DilithiumPKHash`, which encodes to

```cpp
CScript() << OP_DUP << OP_HASH160 << ToByteVector(keyID) << OP_EQUALVERIFY << OP_CHECKSIGDILITHIUM
```

Two separate things are wrong with handing that out.

**It is not a valid payment destination on most chains.** `ValidDestinationVisitor` gates it behind

```cpp
bool LegacyDilithiumBase58PaymentsAllowed()
{
    return Params().GetConsensus().nDilithiumP2MRHeight == std::numeric_limits<int>::max();
}
```

Only the live testnet leaves `nDilithiumP2MRHeight` unscheduled. Mainnet, signet and regtest all set it to 1, so `IsValidDestination(DilithiumPKHash)` is false on all three -- while `GetNewDestination` still returns one. The failing test runs on mainnet parameters, so this is observed rather than inferred.

**The wallet cannot spend it either.** `CalculateMaximumSignedInputSize` returns -1 for the script:

- `InferDescriptor` cannot model a Dilithium P2PKH, so the descriptor path yields nothing.
- The `DummySignInputWeight` fallback added for P2MR only accepts witness satisfactions -- it returns `nullopt` whenever `sigdata.scriptSig` is non-empty, which is always true for a legacy spend.

`AvailableCoins` derives `solvable` directly from that value (`bool solvable = input_bytes > -1;`), so any such output is unsolvable and coin selection can never include it. The -1 also propagates into `COutput::input_bytes`.

## Why this matters

This is the same hazard that `dilithium-bech32` was disabled for, and that one fails loudly:

```
Error: dilithium-bech32 is disabled because witness v0 keyhash programs are
indistinguishable from ECDSA P2WPKH and are not spendable with Dilithium keys
```

`dilithium-legacy` has no equivalent guard, so on a P2MR-only chain it can produce an address that neither consensus nor the wallet will let you spend from.

## Questions for the fix

[...]

Issue #90, opened 2026-07-30
security: Dilithium message signing has no domain separation from transaction signing


BarneyChambers:
## Summary

`signmessagewithdilithium` signs the caller's message bytes directly, with no domain-separation prefix and an empty Dilithium context. Transaction signing uses the *same* key and the *same* primitive over a bare 32-byte sighash. As a result, a message signature and a transaction spending signature are drawn from the same domain, and `signmessagewithdilithium` can act as a signing oracle for the key that spends a P2MR output.

This is pre-existing behaviour on `master`. It was noticed while reviewing #84, which refactors the adjacent verification RPCs but does not introduce or change this.

## Why the two domains collide

Transaction signing reduces to `SignMessage` over the raw sighash bytes:

```cpp
// src/crypto/dilithium_key.cpp
bool CDilithiumKey::Sign(const uint256& hash, std::vector<unsigned char>& vchSig,
                         const std::vector<unsigned char>& context) const
{
    if (!IsValid()) return false;
    // Use the hash as the message to sign
    return SignMessage(Span<const unsigned char>(hash.begin(), hash.size()), vchSig, context);
}
```

`MutableTransactionSignatureCreator::CreateDilithiumSig` calls it with the default (empty) context:

```cpp
// src/script/sign.cpp
if (!SignatureHashSchnorr(hash, execdata, m_txto, nIn, static_cast<uint8_t>(hashtype),
                          sigversion, *m_txdata, MissingDataBehavior::FAIL)) return false;
if (!key.Sign(hash, vchSig)) return false;
vchSig.push_back((unsigned char)hashtype);
```

The message RPC signs user bytes through the same call, also with the default empty context:

```cpp
// src/wallet/rpc/dilithium.cpp
std::vector<unsigned char> messageBytes(strMessage.begin(), strMessage.end());
if (!dilithium_key.SignMessage(messageBytes, vchSig)) { ... }
```

Both `Sign` and `SignMessage` default `context` to `{}`, so nothing distinguishes the two uses. Contrast this with the inherited ECDSA `signmessage`, which commits to the `"Bitcoin Signed Message:\n"` magic and a length prefix before hashing, precisely so that a message signature can never be a transaction signature.

## Why the key is the same key

[...]

Issue #85, opened 2026-07-27
psbt: P2MR spend data cannot cross devices — no PSBT fields for merkle path, leaf script or Dilithium key material


augchan42:
Split out of #79 so that PR #83 is not mistaken for closing the whole problem.

## What #83 does and does not fix

#83 makes `walletprocesspsbt` sign P2MR inputs **when the signing wallet is the wallet that holds the P2MR metadata**. That covers the reported case, because `CWallet` can read the merkle path, leaf script and Dilithium key from its own store — the same way `signrawtransactionwithwallet` already did.

It does nothing for the case where the signer is a *different* device.

## The remaining gap

The PSBT serialisation carries no fields for P2MR spend data. A PSBT handed to another signer contains only `witness_utxo` for a P2MR input:

```
"inputs": [ { "witness_utxo": {
  "scriptPubKey": { "type": "witness_v2_p2mr",
    "hex": "5220747faea21cc9c03ef488b7c58c255c70a0989652c09c44cebd5c9419a06c5dfa" } } } ]
```

No merkle path, no leaf script, no key origin. A signer without the originating wallet's metadata has nothing to work from, and no way to report why.

## Why it matters

This is the workflow every custodian and hardware wallet uses: an online coordinator builds the PSBT, an offline or hardware signer signs it. Until P2MR data can travel in the PSBT, quantum-safe coins cannot participate in any air-gapped or multi-device flow — which is the flagship output type of the chain.

Not urgent for testnet, where the single-wallet path now works. Squarely mainnet-relevant.

## Direction

Define PSBT key types for the P2MR spend data — proprietary to start, or adopt BIP-360's fields if and when they are specified. Populate them in `walletcreatefundedpsbt`, consume them in the input signer. Worth checking whether the taproot precedent (`PSBT_IN_TAP_LEAF_SCRIPT`, `PSBT_IN_TAP_BIP32_DERIVATION`) can be mirrored rather than invented from scratch, since P2MR is structurally close to tapscript.

Related: #79 (single-wallet case, fixed by #83), #32 work item 3.

Issue #76, opened 2026-07-21
default change output address type is TapRoot (even sending dilithium to dilithium)?


unkn0wn747:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behaviour

When sending coins between two dilithium addresses, the change goes to myself at TapRoot address type.
With time it may silently migrate more and more coins to non-QR (TR) address type. 

### Expected behaviour

Dilithium should be the default type address for change return outputs in all type of TXs.

### Steps to reproduce

Send coins between addresses, see outputs in TX, the change goes to TapRoot address type.

### Relevant log output

https://explorer.bitcoinquantum.com/tx/fa73ccb99853142995cf28a64b63aedea50c96fb0b66f28d9694c360ed4c55cc

### How did you obtain BTQ Core

Compiled from source

### What version of BTQ Core are you using?

v0.4.1

### Operating system and version

Ubuntu 24

BarneyChambers:
Why change is Taproot
Wallet change selection still follows Bitcoin Core–style defaults: if the wallet can create Taproot (bech32m) change, it prefers that. Dilithium destinations are not considered when choosing the change address type. So even Dilithium → Dilithium sends can leave change on a non-QR Taproot address.

I think you're right that this isn't ideal, so we will consider a fix for this now. Thank you for the finding!

unkn0wn747:
### _Root cause analysis by Claude Opus 4.8 (from master, line numbers as of today)_

Thanks for looking at this. I dug into the code, and there appear to be two independent blockers, so a fix limited to change-type detection alone will not be sufficient.

### 1. Dilithium destinations are invisible to change-type selection

CWallet::TransactionChangeType() (src/wallet/wallet.cpp:2225-2284) inspects only four destination variants:

cpp
for (const auto& recipient : vecSend) {
    if (std::get_if<WitnessV1Taproot>(&recipient.dest))      any_tr = true;
    else if (std::get_if<WitnessV0KeyHash>(&recipient.dest)) any_wpkh = true;
    else if (std::get_if<ScriptHash>(&recipient.dest))       any_sh = true;
    else if (std::get_if<PKHash>(&recipient.dest))           any_pkh = true;
}

DilithiumPKHash (P2DPKH) and WitnessV2P2MR (P2MR) are never examined, so no branch can ever return a Dilithium change type. Execution then reaches the fallback:

cpp
if (has_bech32m_spkman) {
    return OutputType::BECH32M;
}

[...]

PR #101, opened 2026-07-30
doc: add external audit briefing

BarneyChambers:
Adds `doc/audit-briefing.md`, orientation material for the external review scheduled against this repository, and links it from the docs index.

## What it contains

- **The audit commit.** `5ce8e10bdc9d0b1e8782ec35d9d5b239df7f8ac3`, the merge of #96. It is the first commit at which the test suites can be used as a signal at all, and everything the document measures was measured there.
- **The divergence tables.** Thirty consensus, policy and encoding constants that differ from Bitcoin Core, each with the file it lives in. A few unchanged values are listed on purpose, because some of the sharper effects come from a constant that moved sitting next to one that did not — `WITNESS_SCALE_FACTOR` at 16 against `MAX_STANDARD_TX_WEIGHT` still at 400,000 being the clearest case.
- **The interactions.** Consequences that fall out of the tables rather than from any line of code: non-witness block capacity halving while witness capacity doubled, `MAX_PACKAGE_WEIGHT` being unreachable under the count limit, per-input weight tripling what a wallet pays to consolidate, and every soft fork being active from height 1.
- **Where the review is most valuable.** Ranked by how much of the code is new rather than inherited, on the reasoning that upstream code has had far more scrutiny than anything written for this fork.
- **Open items and test-suite scope**, including which suites are green, how that was measured, and which are not.

## Why the framing is what it is

Two things a reviewer would otherwise have to discover the hard way are stated up front.

The first is that a green suite here is recent. Until #96 the unit suite could not run to completion, `make check` silently skipped most of `wallet/test` and all 95 `btq-tx` cases, and that case table invoked binaries this fork does not build. The document says so and advises breaking a rule deliberately rather than trusting that a passing test was ever exercised.

The second is that the functional suite is not green and was not addressed. It is failing for the same reason the unit suite was — inherited constants that no longer describe this chain, such as `rpc_blockchain.py` asserting a UTXO total derived from a 50 BTC subsidy — but until it is migrated, its output cannot distinguish a stale expectation from a defect. Better for the auditors to have that stated than to infer it.

Documentation only; no code changes.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -52,6 +52,7 @@ Development
 The BTQ repo's [root README](/README.md) contains relevant information on the development process and automated testing.
 
 - [Developer Notes](developer-notes.md)
+- [External Audit Briefing](audit-briefing.md)
 - [Productivity Notes](productivity.md)
 - [Release Process](release-process.md)
 - [Source Code Documentation (External Link)](https://doxygen.btqcore.org/)
--- a/doc/audit-briefing.md
+++ b/doc/audit-briefing.md
@@ -209,19 +209,22 @@ for this fork.
 | `src/test/test_btq` | 733 cases, no errors | Four consecutive full runs |
 | `test/util/data` | 95 of 95 cases | `test/util/test_runner.py`, via `make check` |
 | `src/qt/test/test_btq-qt` | 21 cases, no failures | Built with `--with-gui=qt5`, run headless |
-| `test/functional` | Not green, and not addressed | Sampled, not migrated |
+| `test/functional` | 177 of 304 pass, 127 fail | One full sequential run, 25 minutes |
 
-The functional suite is the known gap. It fails for the same reason the unit
-suite did: inherited constants that no longer describe this chain.
-`rpc_blockchain.py` asserts a UTXO total of 8725 BTC, which is Bitcoin's
-50-per-block subsidy over the test chain, against the 1000 that BTQ's subsidy
-produces. That is a stale expectation rather than a defect, but until the suite
-is migrated it cannot distinguish between the two, and neither can a reviewer
-reading its output.
+The functional suite is the known gap, and the 127 failures have **not** been
+triaged. Do not read them as 127 defects, and do not read them as 127 stale
+expectations either; nobody has looked.
 
-Run it with a single job. Under parallel jobs most tests fail within a second on
-node startup contention, which produces a failure list that has nothing to do
-with the code.
+What is known is that the suite fails for at least some of the same reasons the
+unit suite did. `rpc_blockchain.py` asserts a UTXO total of 8725 BTC, which is
+Bitcoin's 50-per-block subsidy over the test chain, against the 1000 that BTQ's
+subsidy produces. That particular one is a stale expectation rather than a
+defect. Until the suite is migrated it cannot distinguish between the two
+categories, and neither can a reviewer reading its output.
+
+Run it with a single job. Under parallel jobs almost everything fails within a
+second on node startup contention, producing a failure list that has nothing to
+do with the code — the run above took 25 minutes at `--jobs=1`.
 
 Two further notes on the tests:
 
--- a/doc/audit-briefing.md
+++ b/doc/audit-briefing.md
@@ -134,13 +134,24 @@ legacy non-witness Dilithium spend puts its signature in the scriptSig at 16
 weight units per byte, four times the upstream penalty, which caps such a
 transaction at about seven inputs.
 
-### The package weight limit may be unreachable
+### The mempool's size limits are now larger than a block
 
-`MAX_PACKAGE_WEIGHT` is 40,000,000 and `MAX_PACKAGE_COUNT` is 25. At the
-standard transaction ceiling of 400,000 weight units, twenty-five transactions
-total 10,000,000, a quarter of the limit. Only non-standard transactions can
-reach it, so in ordinary operation the count limit always binds first and the
-weight limit is close to dead code.
+A block holds 8,000,000 weight units, which at a scale factor of 16 is 500,000
+vbytes. Against that, `DEFAULT_ANCESTOR_SIZE_LIMIT_KVB` and
+`DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` are 2,000 kvB, four times a block, and
+`MAX_PACKAGE_WEIGHT` is 40,000,000, five times `MAX_BLOCK_WEIGHT`. Upstream
+holds all three near a tenth of block capacity, which is what makes an accepted
+package always minable in one block. That ordering is inverted here.
+
+The 40,000,000 looks derived rather than chosen: `policy/packages.h` asserts
+`MAX_PACKAGE_WEIGHT >= DEFAULT_ANCESTOR_SIZE_LIMIT_KVB * WITNESS_SCALE_FACTOR *
+1000`, so the 2,000 kvB ancestor limit forced it. The ancestor limit is
+therefore the value to ask about.
+
+Separately, `MAX_PACKAGE_COUNT` is still 25, and twenty-five transactions at the
+standard ceiling of 400,000 weight units total 10,000,000 — a quarter of the
+package weight limit. In ordinary operation the count limit always binds first
+and the weight limit is close to dead code.
 
 ### Per-input weight changes what a wallet can spend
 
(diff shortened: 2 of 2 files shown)


augchan42 (approved):
Read it in full and verified the divergence tables against source at `5ce8e10bd` rather than reading them. **Every one of the constants checks out** — weight and size, script, mempool policy, chain, encoding. I could not find a wrong value.

That is worth stating explicitly given how much of today involved catching numbers that were confidently stated and wrong, including several of mine.

The framing is also right in a way that matters: leading with divergences rather than code, and telling the reviewer that a green suite here is *recent* and to prefer breaking a rule deliberately to confirm the test catches it. That is the single most useful sentence in the document.

Four things below. Only the first two change anything.

## 1. Two of the "constants" are derived, not independent

`MAX_STANDARD_SCRIPTSIG_SIZE` and `MAX_STANDARD_P2WSH_STACK_ITEM_SIZE` are not separately chosen values:

```cpp
// policy.h:46
static constexpr unsigned int MAX_STANDARD_P2WSH_STACK_ITEM_SIZE{MAX_SCRIPT_ELEMENT_SIZE};
// policy.h:52
static constexpr unsigned int MAX_STANDARD_SCRIPTSIG_SIZE{MAX_SCRIPT_ELEMENT_SIZE};
```

Both are `MAX_SCRIPT_ELEMENT_SIZE`. The table lists three raised limits where the tree has **one raised limit and two that followed it**.

This strengthens the document's own thesis rather than weakening it: raising `MAX_SCRIPT_ELEMENT_SIZE` from 520 to 15,000 silently moved two mempool policy limits by a factor of 29 and 187 respectively, and nothing records that as a decision. It is exactly the "constant that moved sitting next to one that did not" point, one level deeper.

Suggest marking them as derived in the table. A reviewer who sees three independent numbers may go looking for three rationales.

There is at least a guard: `static_assert(MAX_STANDARD_SCRIPTSIG_SIZE * WITNESS_SCALE_FACTOR < MAX_STANDARD_TX_WEIGHT)` — 15,000 × 16 = 240,000 against 400,000, so it holds, but with less headroom than the numbers suggest in isolation.

## 2. `MESSAGE_MAGIC` is listed as changed, but the Dilithium path never uses it

```
$ grep -rn 'MESSAGE_MAGIC' src/wallet/rpc/dilithium.cpp src/crypto/dilithium_key.cpp
(no matches)
```

The table entry is correct for the ECDSA path (`util/message.cpp:76`). But a reviewer reading "MESSAGE_MAGIC: `BTQ Signed Message:\n`" in a divergence table will reasonably conclude BTQ message signing is domain-separated. It is for ECDSA and is not for Dilithium — which is #90, where `signmessagewithdilithium` signs raw bytes with an empty context over the same key and primitive as transaction signing.

Two suggestions:

- Note in that row that it applies to the ECDSA path only.
- **#90 is missing from "Known open items"**, which lists #97, #94 and #95. It is the most severe item open — a spending oracle on the chain's headline feature — and its absence from the list a reviewer will use as their index seems more likely to be an oversight than a judgement.

Related: "Domain separation between message signing and transaction signing has already required one fix" reads as though #90 is resolved. It is open. Worth rewording unless it refers to something else.

## 3. Your functional numbers and mine agree, independently

You measured 177 pass / 127 fail, serial, 25 minutes. I measured 162 pass / 126 fail / 16 skipped at the same commit, `--jobs 8`, 3.6 minutes, with #106 applied. Counting your figure as including skips, that is agreement to within one test from two different methods.

That is useful for the audit: the functional baseline is reproducible, which is more than could be said of anything in this repo yesterday.

## 4. The single-job caveat has a cause, and a fix

> Run it with a single job. Under parallel jobs almost everything fails within a second on node startup contention

Correct, and the cause is #105: `p2p_port()` and `rpc_port()` derive ports from the node index alone, so `PortSeed` is dead code and every concurrent job binds the same ports. Since `--jobs` defaults to 4, the *default* invocation produces ~261 spurious failures.

#106 restores the seed term. Once it lands, that paragraph can become "run it however you like", and the 25-minute serial run becomes about four minutes — which is the difference between a suite CI can gate on and one it cannot.

## Nits

- The closing CI note — *"a green tick on `master` is a narrower claim than it looks"* — is the kind of thing auditors rarely get told voluntarily. Worth keeping prominent.
- Worth stating the Dilithium parameter set explicitly somewhere (Dilithium2: 1312-byte keys, 2420-byte signatures). It is load-bearing for the weight arithmetic in the interactions section, and a reviewer should not have to derive it.

Approving on the substance — the tables are accurate and the interactions section is the most valuable part. The two items above are worth folding in before an external reader sees it.

augchan42:
**Following up on my own approval.** I approved this before I had the functional baseline and before checking `nLWMAHeight`, and four things in it are now wrong or incomplete. The document is good — it is the best orientation material in the repo and the interaction analysis is the part an auditor will actually use — which is why I would rather it be exact.

Verified correct, so nobody re-checks them: the `nDilithiumP2MRHeight` row (1 on mainnet/signet/regtest, `INT_MAX` on testnet), `nSubsidyHalvingInterval` 2,100,000, mainnet magic `f1b2a3d4`, and all the weight arithmetic — 768 vs 272 per P2WPKH input, ~519 vs ~1,470 inputs, 25,000 vs 100,000 base bytes.

---

### 1. The LWMA row omits its activation height, and that changes what it means

| network | `nLWMAHeight` |
| -- | -- |
| **mainnet** | **1** |
| testnet | 300000 |
| signet | 300000 |
| regtest | 300000 |

The divergence table says "LWMA, window 45, every block" with no height, and §"Where the review is most valuable" ranks it **third** with "replaces the 2016-block retarget entirely and runs every block against 60-second spacing." An auditor reads that as a live, exercised algorithm.

It is not. Testnet is around 200k and still on the legacy retarget — it activates at 300,000, currently estimated **Aug 10-15**. So **LWMA has never run on a live chain**, and because regtest also gates at 300,000 while functional chains are ~200 blocks, **no functional test exercises it either**. The entire evidence base is the 12 synthetic unit cases in `pow_lwma_tests.cpp`.

This also cuts against §"Every soft fork is active from height 1", which is otherwise accurate. LWMA is the one deployment that runs the *opposite* way: active from 1 on mainnet and nowhere else. That inversion is worth stating explicitly, because it means mainnet is the first place this code runs in anger.

I filed #110 on the related question — `LWMA_WINDOW = 45` carries no justifying comment anywhere in the tree, and at 60s spacing that window spans 45 minutes of history where the same constant gives a Bitcoin-spaced chain 7.5 hours.

### 2. The functional failures have now been triaged

> the 127 failures have **not** been triaged ... nobody has looked.

Someone has now. Baselined at the audit commit: **162 pass / 126 fail / 16 skip**. Your 177/127 looks like skips counted as passes — same 304 total, so we agree on the measurement.

The dominant cause is identified: BTQ pays 5 BTQ per block where upstream pays 50, and inherited tests hardcode the upstream figure. Roughly half the failures are arithmetic, not behaviour. Your `rpc_blockchain.py` 8725-vs-1000 example is exactly this cluster. Tracked in #104 with the sweep pattern.

### 3. "Run it with a single job" documents a bug as a characteristic

> Under parallel jobs almost everything fails within a second on node startup contention

That is #105, and the fix is open in #106. `p2p_port()`/`rpc_port()` derive ports from the node index alone and ignore `PortSeed`, so every concurrent job binds identical ports. Since `--jobs` defaults to 4, the *default invocation* produces ~261 spurious failures.

As written, the briefing tells an external auditor to work around a defect we have a one-file fix for, and implies the suite is inherently serial. It is not — this is a dropped upstream term. Worth reframing as a known bug with a pending fix, or landing #106 first and deleting the paragraph.

### 4. "Known open items" omits two of the three Urgent issues

Listed: #97, #94, #95. Missing:

- **#90** (Urgent) — Dilithium message signing has no domain separation from transaction signing. §"Where the review is most valuable" says this "has already required one fix," which reads as resolved; the issue is open.
- **#76** (Urgent) — default change output is Taproot even when spending Dilithium, silently eroding the quantum-safety property that is the entire point of the chain.
- **#103** — Dilithium has no Known Answer Tests. For a post-quantum chain this is close to the first thing a cryptography auditor asks for, and the briefing ranks Dilithium as the #1 review target without mentioning that its primitive is unvalidated against published vectors.
- **#85** — P2MR spend data cannot cross devices in a PSBT; currently mainnet-blocking. Design under review in #99.

Omitting known Urgent issues from an auditor briefing costs more than it saves — if they find #90 or #76 independently, every other claim in this document gets re-checked.

---

Happy to push these edits myself if that is faster. On CI: the `test each commit` failure here is real rather than stale — your branch is correctly based on post-#96 master, so unlike #99 and #100 it will not clear on a rebase.

Issue #104, opened 2026-07-30
test: 126 functional tests fail — dominant cause is the 5-vs-50 block subsidy


augchan42:
## Summary

The unit suite is now green — #96 gets `test_btq` to 133 suites / 733 cases / 0 failures. **The functional runner has not been measured at all.**

`doc`-level definition of done for the test work is "`make check` green **and** functional runner green". Only the first half is demonstrated.

## What is actually known

Three functional failures were recorded on 2026-07-27, verified pre-existing on unmodified master at the time by re-running against a stashed working tree:

- `rpc_help.py` — `RPC client conversion table (src/rpc/client.cpp) and RPC server named arguments mismatch!`
- `wallet_basic.py` — `not(5.00000000 == 50)`, looks like an inherited block-subsidy expectation
- `rpc_psbt.py` — `not(19 == 2)`

**Nobody has re-run these since**, and master has moved substantially: #83, #87, #88, #89, #91 merged, #96 in flight. The counts should be treated as unverified rather than small.

That caution is earned. The unit-side estimates from the same measurement session turned out to be wrong in both directions — `miner_tests` was recorded as 112 independent failures and proved to be three root causes, while several suites had never executed at all and were invisible.

## `rpc_help.py` deserves attention regardless of count

It is a real consistency check across the whole RPC surface, and while it is red it silently absorbs genuine mistakes. It could not, for example, verify the new RPC added in #84. **Fix rather than skip.**

## Suggested approach

1. Build at CI parity and run `test/functional/test_runner.py` to get an honest current baseline.
2. Triage each failure on fix-vs-skip merits individually — none should be batch-disabled.
3. Record any skip with a rationale, per the policy on the test umbrella.

Worth checking activation-height assumptions first in anything that looks like stale data. Every fatal in the unit chain reduced to one shape: **BTQ activates at height 1 what Bitcoin activates at a fork height** (`SegwitHeight` 1 vs 481824, `CSVHeight` 1 vs 419328, LWMA from block 1). Code paths dormant across an inherited test's heights upstream are live here. `wallet_basic.py`'s subsidy expectation may well be simple stale data, but the others are worth checking against that pattern before assuming so.

---

Tracked in Linear as [BTQ-115](https://linear.app/btq/issue/BTQ-115). Related: #60.

augchan42:
## Functional baseline — first measurement, 2026-07-30

Run against `upstream/master` at `cc1479494` (with #91 merged), built at CI parity.

[...]

PR #106, opened 2026-07-30
test: restore PortSeed in functional port allocation

### LINKED ISSUE #105: test: functional suite cannot run in parallel — port allocation ignores PortSeed
## Summary

`test/functional/test_framework/util.py` derives node ports from the node index alone, ignoring `PortSeed`. As a result **every parallel test job binds the same ports**, and any run with `--jobs > 1` produces mass spurious failures.

```python
def p2p_port(n):
    # BTQ: Use BTQ-specific ports for BTQ regtest
    return 19444 + n

def rpc_port(n):
    # BTQ: Use BTQ-specific ports for BTQ regtest
    return 18443 + (n * 10)
```

Upstream derives them as `PORT_MIN + n + (MAX_NODES * PortSeed.n) % (PORT_RANGE - 1 - MAX_NODES)`. The `PortSeed.n` term is exactly what gives concurrent jobs disjoint ranges, and it has been dropped.

`PortSeed` is now dead code:

- declared at `util.py:307`
- assigned at `test_framework.py:228` — `PortSeed.n = self.options.port_seed`
- **never read anywhere**

So `test_runner.py` computes a distinct `--portseed` per job (line 719), passes it, the framework stores it, and nothing consumes it. Two concurrent jobs both put node 0 on rpc 18443 and p2p 19444.

## Symptom

```
btqd exited with status 1 during initialization.
Error: Unable to start HTTP server. See debug log for details.
AssertionError: [node 0] Error: no RPC connection
```

## Measured on master `cc1479494`

| invocation | result |
|---|---|
| `--jobs 8` | 276 of 304 "failed" in 118 s |
| `--jobs 4` | 261 of 304 "failed" in 122 s |
| `--jobs 2` | 4 of 6 probe tests failed |
| serial | those same 4 tests **all pass** |

The failing set differs between runs, which is the tell for a race rather than a defect in the tests.

## Why this matters beyond inconvenience

**Any functional failure count taken from a parallel run is noise.** I nearly recorded "276 failures" as this repo's functional baseline. Someone acting on such a number would either conclude the suite is catastrophically broken, or spend days "fixing" tests that were never broken.

It also means CI cannot run functional tests in parallel, and a serial run of 304 tests is slow enough that it is likely to be skipped — so in practice the functional suite is not being run at all.

## Fix

Restore the `PortSeed.n` term in both functions, keeping BTQ's base ports:

```python
def p2p_port(n):
    assert n <= MAX_NODES
    return PORT_MIN + n + (MAX_NODES * PortSeed.n) % (PORT_RANGE - 1 - MAX_NODES)
```

Worth checking why the BTQ-specific bases were introduced at all — regtest ports are already namespaced by `chainparamsbase.cpp`, and the framework's `PORT_MIN` default of 11000 does not collide with them. If the change was working around a collision with a locally-running node, the portseed mechanism already solves that more generally.

Related: #104 (functional suite has never been baselined) — this is why.

augchan42:
## Summary

Fixes #105. The functional suite could not be run in parallel — and since `--jobs` defaults to 4, that means it could not be run *at all* in its default invocation without producing near-total spurious failure.

`p2p_port()` and `rpc_port()` derived ports from the node index alone:

```python
def p2p_port(n):
    return 19444 + n

def rpc_port(n):
    return 18443 + (n * 10)
```

So every concurrent job bound the same ports. Job B's node 0 collided with job A's node 0 and died during initialisation:

```
btqd exited with status 1 during initialization.
Error: Unable to start HTTP server.
AssertionError: [node 0] Error: no RPC connection
```

`PortSeed` had become dead code as a consequence — `test_runner.py` computes a distinct `--portseed` per job and passes it (line 719), `test_framework.py:228` stores it in `PortSeed.n`, and **nothing read it**. That term is exactly what turns a per-job integer into a disjoint port range.

## Measured on `cc1479494`

| invocation | before | after |
|---|---|---|
| `--jobs 8` | 276 of 304 "failed" in 118 s | see below |
| `--jobs 4` | 261 of 304 "failed" in 122 s | 6-test probe passes cleanly |
| `--jobs 2` | 4 of 6 probe tests failed | — |
| serial | those same 4 tests pass | — |

The failing set differed between runs, which is the signature of a race rather than of broken tests. `example_test.py` — the trivial reference test — was among the "failures", and 304 tests "completing" in 118 seconds is not a test run, it is 304 nodes failing to start.

Port ranges are now provably disjoint across jobs:

```
seed 1: p2p 11012-11014, rpc 16012-16014
seed 2: p2p 11024-11026, rpc 16024-16026
seed 3: p2p 11036-11038, rpc 16036-16038
seed 4: p2p 11048-11050, rpc 16048-16050
96 ports across 4 jobs, 96 distinct -> disjoint
```

## Why this matters beyond convenience

**This is very likely why the functional suite has never been baselined** (#104). Anyone who ran it saw near-total failure and reasonably concluded the tree was broken, or started "fixing" tests that were never broken.

It also means **any functional failure count taken from a parallel run is noise**. I nearly recorded 276 as this repo's functional baseline before running the canary check.

## Approach

Restores upstream's formula rather than inventing a new scheme, per minimal-change principles — the divergence introduced here was dropping the seed, so the fix is putting it back.

The BTQ-specific bases were not needed. Each node is told its ports explicitly by `initialize_datadir()`, which writes `port=` and `rpcport=` into that node's `btq.conf` (`util.py:399-400`), so these values need not match the regtest defaults in `chainparamsbase.cpp`. Nothing else in `test/functional/` hardcodes 18443 or 19444 — checked.

If the range ever collides with something running locally, `TEST_RUNNER_PORT_MIN` shifts it. Dropping the seed is not the way to solve that, and it costs parallel execution entirely.

## Test plan

- [x] 6-test probe at `--jobs 4`: was 4 of 6 failing, now 6 of 6 pass
- [x] Port allocation verified disjoint across 4 concurrent seeds
- [x] Confirmed nothing else in `test/functional/` depends on the 18443/19444 bases
- [x] Full suite re-run at `--jobs 8` — the setting that originally produced 276 failures
- [ ] The genuine functional failures this exposes are a separate matter, tracked in #104

## Note

This unblocks #104. A real functional baseline was not achievable while the runner's default invocation produced 261 false failures.

### DIFF
--- a/test/functional/test_framework/util.py
+++ b/test/functional/test_framework/util.py
@@ -334,22 +334,28 @@ def get_rpc_proxy(url: str, node_number: int, *, timeout: Optional[int]=None, co
     return coverage.AuthServiceProxyWrapper(proxy, url, coverage_logfile)
 
 
+# Both of these must include the PortSeed.n term. test_runner.py assigns every
+# parallel job a distinct --portseed, and that term is what turns it into a
+# disjoint port range per job. Deriving ports from the node index alone makes
+# every concurrent job bind the same ports, so node 0 of job B collides with
+# node 0 of job A and dies with "Unable to start HTTP server". Because --jobs
+# defaults to 4, that is the *default* invocation, not an opt-in: it produced
+# ~261 spurious failures out of 304 before this was restored, while the same
+# tests pass serially.
+#
+# These do not need to match the regtest defaults in chainparamsbase.cpp. Each
+# node is told its ports explicitly via initialize_datadir(), which writes
+# port= and rpcport= into that node's btq.conf.
+#
+# If this range ever collides with something else running locally, shift it with
+# the TEST_RUNNER_PORT_MIN environment variable rather than by dropping the seed.
 def p2p_port(n):
     assert n <= MAX_NODES
-    # BTQ: Use BTQ-specific ports for BTQ regtest
-    # BTQ regtest P2P port: 19444, RPC port: 18443
-    # For multiple nodes, increment by 1 for each node
-    # Use a larger range to avoid conflicts between P2P and RPC ports
-    return 19444 + n
+    return PORT_MIN + n + (MAX_NODES * PortSeed.n) % (PORT_RANGE - 1 - MAX_NODES)
 
 
 def rpc_port(n):
-    # BTQ: Use BTQ-specific ports for BTQ regtest
-    # BTQ regtest RPC port: 18443 (from chainparamsbase.cpp)
-    # For multiple nodes, increment by 1 for each node
-    # Use a larger range to avoid conflicts between P2P and RPC ports
-    # Start RPC ports at 18443, but use a larger increment to avoid conflicts
-    return 18443 + (n * 10)
+    return PORT_MIN + PORT_RANGE + n + (MAX_NODES * PortSeed.n) % (PORT_RANGE - 1 - MAX_NODES)
 
 
 def rpc_url(datadir, i, chain, rpchost):


augchan42:
Checked the history rather than assuming the hardcoding was arbitrary, and it was not — this needs saying, because the original change was solving a real problem.

`c007c0a0b` (@Oskii, 2025-08-25), *"test-framework(util): BTQ regtest ports, regtest config, and stable startup"*, states its purpose as:

> Avoids HTTP server start failures and port conflicts under multi-node runs.

That is the same symptom this PR is fixing, which is worth understanding before anyone reverts on the grounds that the original fixed something.

**The commit made three changes, not one:**

1. hardcoded `p2p_port` / `rpc_port`, dropping the `PortSeed` term
2. added `regtest` handling in `write_config`
3. stopped writing `connect=0`, because it implicitly sets `listen=0` and conflicts with the `bind=127.0.0.1` the framework adds

**Change 3 is almost certainly what actually fixed the reported symptom.** A node started with `connect=0` gets an implicit `listen=0`, which contradicts the explicit bind — that is a genuine startup failure with a correct fix, and it has nothing to do with port numbering.

Change 1 looks like it was co-diagnosed rather than causal. Under upstream's formula p2p occupies `PORT_MIN..PORT_MIN+PORT_RANGE` and rpc occupies `PORT_MIN+PORT_RANGE..`, so the two can never overlap within a run — the "conflicts between P2P and RPC ports" the comments describe are not reachable with the seed term present.

**This PR reverts only change 1.** Changes 2 and 3 are untouched and remain on the branch — verified: the diff against master is 30 lines, entirely within `p2p_port` and `rpc_port`, and touches nothing in `write_config`.

So the two fixes compose: keep the `connect=0` / `listen=0` fix that made single-test multi-node startup reliable, and restore the seed term that makes concurrent jobs use disjoint ranges. Neither displaces the other.

Issue #114, opened 2026-07-30
launch: mainnet has no working peer discovery — sole DNS seed does not resolve, fixed seeds are a placeholder


augchan42:
A fresh mainnet node currently has **no way to find a peer**. Both discovery mechanisms are non-functional, and the second is explicitly labelled as such in the source.

Found while verifying claims in the auditor infrastructure brief. Not a code defect — a launch-configuration gap — but it is the kind an external auditor leads with.

## 1. The only mainnet DNS seed has no A record

```cpp
// src/kernel/chainparams.cpp — CMainParams
vSeeds.emplace_back("seed1.btq.com");
```

That is the entire mainnet seed list. Bitcoin mainnet ships nine.

Resolution check, control first so this is not mistaken for a sandbox DNS problem:

| host | A record |
| --- | --- |
| `github.com` | 20.205.243.166 |
| `bitcoinquantum.com` | 76.76.21.21 |
| `btq.com` | 216.150.1.1 |
| **`seed1.btq.com`** | **none** |
| **`signet-seed1.btq.com`** | **none** |
| `testnet-seed1.bitcoinquantum.com` | 208.85.16.175 |

The apex domains resolve and the testnet seed resolves. Only the `btq.com` subdomains are absent, so this is missing records rather than filtering.

## 2. The fixed-seed fallback is a placeholder that says so

```cpp
// src/chainparamsseeds.h
static const uint8_t chainparams_seed_main[] = {
    0x01, 0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00  // Placeholder entry (invalid, will be ignored)
};
```

`vFixedSeeds` is the mechanism that is supposed to cover exactly the case where DNS seeding fails. Here both fail together.

Net effect on mainnet: a node started with default configuration cannot bootstrap. It works only if the operator supplies `-addnode`/`-connect` manually.

## 3. The question underneath: do we control `btq.com`?

This is the part I cannot answer from the repository and it should be answered today.

Every other piece of BTQ infrastructure sits on `bitcoinquantum.com` — the website (Vercel, 76.76.21.21), docs, explorer, pool, and the working testnet seed. Team email is `@btq.tech`. Meanwhile `btq.com` resolves to 216.150.1.1, which does not match any known BTQ infrastructure.

**DNS seeds are a trust anchor.** Whoever controls the seed domain decides which peers a brand-new node meets before it has any chain to validate against. If `btq.com` is not ours:

[...]

Issue #110, opened 2026-07-30
consensus: LWMA_WINDOW=45 is an inherited constant unexamined against BTQ's 60s block spacing


augchan42:
> **Correction 2026-08-02 — the "no empirical data" premise is retired.**
> @unkn0wn747 reported a difficulty cliff observed on the **pre-reset v0.3.2 testnet chain** in April 2026, above height 306,000. That release set `nLWMAHeight = 300000`, so **LWMA was active** — and the magnitude confirms it independently: ~100 → ~200,000 in 2-3 hours is impossible under the legacy retarget, which clamps at 4x per adjustment every 20,160 blocks (six retargets ≈ 120,960 blocks).
>
> So the statement below that **"LWMA has never run on a live chain" is false.** It ran, and it produced a cliff: ~2-3 hours to climb, ~24 hours to recover — an 8-10x asymmetry on the setting chosen for responsiveness. That is evidence *against* N=45, not merely an absence of evidence for it.
> Discussion, the weighting arithmetic and a measured N sweep are in the comments.
> The constant quoted below is now a `Consensus::Params` field in PR #124 (value unchanged), so it can be swept in tests.

Raised during a triage re-check. Tracked in Linear as BTQ-43, which has existed since April carrying the note that the LWMA algorithm's original author suggested revisiting `N=45` in favour of `N=144` or `288` ([zawy12/difficulty-algorithms#50](https://github.com/zawy12/difficulty-algorithms/issues/50)). It had no GitHub issue and no priority.

## The constant

```cpp
// src/consensus/params.h:143
static constexpr int LWMA_WINDOW = 45;
```

No comment, no reference, no derivation — in the whole tree this is the only mention outside `pow.cpp` and its tests. Nothing in `doc/` or `doc-btq/` justifies it.

## Why 45 does not mean on BTQ what it means upstream

`N` is a count of blocks, so its meaning in wall-clock time is set by block spacing, and **BTQ diverges from Bitcoin by 10x here** (`nPowTargetSpacing = 1 * 60`):

| | spacing | N=45 covers |
| -- | -- | -- |
| Bitcoin-spaced chain | 600s | **7.5 hours** |
| BTQ | 60s | **45 minutes** |

[...]

Issue #103, opened 2026-07-30
crypto: Dilithium has no Known Answer Tests — the vendored test vectors are absent


augchan42:
## Summary

`src/crypto/dilithium/SHA256SUMS` contains exactly three entries:

```
tvecs2   tvecs3   tvecs5
```

**None of those files exist in the tree.** `runtests.sh` and `runlcov.sh` were vendored; the test vectors they consume were not. The manifest verifies nothing, and the upstream CRYSTALS-Dilithium validation harness cannot be run against our copy.

**BTQ has no Known Answer Tests for Dilithium.**

## Why round-trip tests do not close this

`dilithium_key_tests` signs and then verifies. That catches gross breakage — a corrupted build, a broken serialisation path.

It **cannot** catch a self-consistent but subtly incorrect implementation, because such an implementation passes a round-trip test perfectly. If our Dilithium diverges from the specification in some corner — a rejection-sampling bound, a packing detail, a hint-encoding edge case — every test we currently run still passes. The defect surfaces only when a BTQ signature has to be verified by something that is not our own binary, or when an attacker finds the divergence first.

For a chain whose entire value proposition is post-quantum signatures, "our implementation agrees with itself" is not the assurance to take into an external audit.

## A prior question: which specification are we claiming?

Worth settling before choosing vectors, because it determines what the test proves.

The vendored `ref/` implementation carries **post-round-3 markers**: `TRBYTES 64`, `RNDBYTES 32`, `CTILDEBYTES` varying 32/48/64 by level, and the API takes a context string:

```c
int pqcrystals_dilithium2_ref_verify(const uint8_t *sig, size_t siglen,
                                     const uint8_t *m, size_t mlen,
                                     const uint8_t *ctx, size_t ctxlen,
                                     const uint8_t *pk);
```

Context strings are a **FIPS 204 / ML-DSA** feature; round-3 Dilithium has none of these. Yet everything is named `Dilithium2` / `pqcrystals_dilithium2_ref`, and the consensus opcodes are `OP_CHECKSIGDILITHIUM`.

So BTQ appears to be running an ML-DSA-aligned implementation under Dilithium naming. That should be confirmed and stated deliberately — it has audit and external-communication consequences beyond this issue.

## Two different tests

| Source | Proves | Cost | Can it fail informatively? |
|---|---|---|---|
| pq-crystals `tvecs2/3/5` | our copy matches the upstream we copied from | low — `SHA256SUMS` is already in-tree as an anchor | no |
[...]

PR #119, opened 2026-07-31
doc: refresh the mainnet audit matrix against current master

augchan42:
Named as a Prerequisite in the external audit SOW. Last revised **2026-06-04**, so it predates seven weeks of work — including the entire unit-suite repair.

Renamed `MAINNET_AUDIT_MATRIX_2026-06-03.md` → `MAINNET_AUDIT_MATRIX.md`. The date implied a snapshot, but the document closes by asking to be updated after each stage, so it was always meant to be living. Git carries the history, and the SOW references it without a date.

## Closed

The `script_tests` validation boundary. Unit suite is **133 suites / 733 cases / 0 failures** at `5ce8e10bd`, CI green on master. The stale-vector buckets named in June — uncompressed P2WPKH expectations, Dilithium opcode reservations, the 15,000-byte push / 100,000-byte script limit drift — were repaired in #91 and #96.

## Opened, and larger than June assumed

- **The functional runner could not run in parallel at all.** `p2p_port()`/`rpc_port()` ignored `PortSeed`, and `--jobs` defaults to 4, so the *default invocation* produced ~261 spurious failures. That is very likely why the suite had never been baselined — anyone who ran it saw near-total failure and reasonably concluded the tree was broken. #105, fix in #106. Baselined now: **162 pass / 126 fail / 16 skip** (#104), dominated by BTQ's 5 BTQ subsidy against inherited tests hardcoding Bitcoin's 50.
- **Mainnet peer discovery does not work** — added to No-Go as launch-blocking. Sole DNS seed `seed1.btq.com` has no A record; `chainparams_seed_main[]` is a placeholder marked invalid in its own comment. #114.
- **LWMA has never executed on a live chain** and no functional test covers it — `nLWMAHeight` is 300000 on testnet, signet and regtest, and 1 on mainnet alone. #110.
- The Dilithium KAT gap is confirmed rather than suspected — the three files named in `SHA256SUMS` are absent. #103.

The stated validation boundary moves from the unit suite to the functional suite, with a note that a green unit suite is narrower than it looks while macOS and Win64 jobs are disabled.

## The part worth arguing about

C01 and C36 **already named the seed gap in June** and the entries were correct. What had not happened in the seven weeks since was anyone going and checking whether the configured hostname resolves.

Every row in this matrix reads identically whether it was examined yesterday or never examined at all. The document is precise about what evidence is required and silent on whether any was gathered. I have noted that a **last-checked column** should be added before an external reviewer sees it, but not added one myself — populating it honestly means someone going row by row, and guessing would be worse than the current absence.

## Not done

I updated the header, the changes section, the validation boundary and the No-Go list. **I did not reassess all 37 component rows** — most depend on work I did not do and cannot evidence. The row-level refresh is still owed, tracked in BTQ-130.

## Test plan

- [x] Every figure reproduced from measurements at `5ce8e10bd` this week
- [x] Seed resolution checked with a control set (`github.com`, `bitcoinquantum.com` and the testnet seed all resolve; the `btq.com` subdomains do not)
- [x] `nLWMAHeight` values read directly from `kernel/chainparams.cpp`
- [ ] Documentation only, no code paths touched


augchan42:
Force-pushed a correction — **the first push was a pure rename with zero content changes.**

`git show --stat` on the original commit:

```
MAINNET_AUDIT_MATRIX_2026-06-03.md => MAINNET_AUDIT_MATRIX.md | 0
1 file changed, 0 insertions(+), 0 deletions(-)
```

So the PR described a full refresh and delivered a filename change. `git add -A <path1> <path2>` staged nothing, because after the `git mv` the second path no longer existed. `gh pr create` printed `Warning: 1 uncommitted change` and I did not act on it.

Now 94 insertions / 12 deletions, matching the description. Nothing else changed.

Do not review this against the earlier diff.

augchan42:
**Added the "Last verified" column** I flagged as missing but did not write. Force-pushed; diff is now 158 insertions / 51 deletions.

I held off originally because *"populating it honestly means going row by row, and guessed dates would be worse than no column."* That was right, so the column is populated conservatively — a row is dated only where the check covered the component's actual state model, not merely where a test touching it happened to pass.

| | rows |
| --- | ---: |
| Verified during the 2026-07-30 – 08-01 sweep | 13 |
| Verified in the June Stage 0/1 tranche, not since | 5 |
| **Never independently verified** | **19** |

**Nineteen of thirty-seven components have never been checked by anyone**, and until now they were typographically indistinguishable from the ones that had. That is not a criticism of the matrix — the rows themselves are good, and the required-evidence levels are the most useful thing in the repo. It is a criticism of the format, which had no way to express "this row is aspirational."

The note above the table says so plainly, and tells a reader to treat a `never` row as **V0 (inventory only)** regardless of what its Required level column claims.

## Why this specific column

C01 and C36 both named *"launch seed policy"* as an open gap in June. Both were correct. What had not happened in the seven weeks since was anyone checking whether the configured hostname resolves — and it does not, on a domain we no longer control (#114). The gap was identified and then sat, looking identical to gaps that had been closed.

A last-verified date makes that visible at a glance. It is the cheapest structural fix available to this document, and it matters most right before an external reviewer reads it — an auditor who sees 19 `never` rows knows exactly where to spend their time, which is worth more to us than a matrix that looks uniformly complete.

## Not done

I have **not** re-assessed the 37 rows themselves. Most depend on work I did not do and cannot evidence. The dates record who checked what and when; the row *contents* are still owed a pass by whoever did the original work — tracked in BTQ-130.

PR #131, opened 2026-08-03
test: add Dilithium2 NIST known-answer tests

Oskii:
## Summary
- Vendor `PQCsignKAT_Dilithium2.rsp` generated from BTQ's production `ref/` build (deterministic signing, not upstream's randomized default).
- Add `dilithium_kat_tests` to `make check`: SHA256 pin of the response file plus verification of all 100 NIST vectors via `pqcrystals_dilithium2_ref_open` and `btq_dilithium_verify`.
- Update `SHA256SUMS` and `Dilithium2_META.yml` to match BTQ's config; add `contrib/devtools/gen-dilithium-kat.sh` to regenerate vectors after `ref/` changes.

Closes BTQ-120 (KAT vendoring item).

## Test plan
- [x] `./src/test/test_btq --run_test=dilithium_kat_tests`
- [ ] CI `make check` on Ubuntu


BarneyChambers:
Closing this as superseded by #129, which was opened for the same issue about two hours earlier and collides directly: both add a file at `src/test/dilithium_kat_tests.cpp` and both rewrite `src/crypto/dilithium/SHA256SUMS`, so only one can land. No reflection on the work — we picked this up in parallel without noticing.

Recording the difference in case any of it is worth pulling across.

The two take different approaches to the same goal. This PR vendors `PQCsignKAT_Dilithium2.rsp` generated from our own `ref/` build and pins its SHA-256. That's an effective regression detector, but it is self-referential: it proves our implementation still agrees with itself, which is the specific gap #103 was filed about ("our implementation agrees with itself is not the assurance to take into an external audit"). It also replaces the three upstream `tvecs` entries in `SHA256SUMS`, which removes the upstream anchor rather than repairing it.

#129 goes at it from the two directions #103 asked for:

- **Supply chain.** It turns out the vendored `SHA256SUMS` was recoverable as an anchor. The manifest had been taken from upstream `ba7ae0b` while the code came from elsewhere, and `ba7ae0b` is the commit that made hedged signing upstream's default — which changes every digest. Our config keeps deterministic signing, so it could never have matched. Corrected to upstream's own published digests for the deterministic configuration, and `contrib/devtools/dilithium-kat.sh` reproduces them from our sources across all three security levels, 10000 iterations each.
- **Conformance.** NIST ACVP ML-DSA-44 vectors run through `btq_dilithium_keypair_from_seed`/`sign`/`verify`, including NIST's negative cases (modified message, commitment, hint, and z). That's an external authority rather than our own output, and it confirms what was previously only inferable from `params.h`: BTQ implements ML-DSA-44 per FIPS 204.

Two things you found independently that #129 agrees with, so they're not lost: the deterministic-vs-hedged mismatch against upstream's default (documented in `src/crypto/dilithium/PROVENANCE.md`, with a test that fails if it's flipped), and that `SHA256SUMS` needed correcting.

One thing this PR does that #129 deliberately does not: editing `Dilithium2_META.yml`. Those are upstream files kept verbatim as part of the vendored snapshot, so #129 records the discrepancy in `PROVENANCE.md` instead of diverging for cosmetics.

If you want the `.rsp` regression pin as well, it composes fine on top of #129 — `ref/nistkat/PQCgenKAT_sign.c` is in-tree and the two don't conflict once #129's file layout is in place. Happy to add it if you think it earns its keep alongside the reference vectors.

PR #115, opened 2026-07-31
doc: move security and governance contacts off a domain we no longer control

augchan42:
Please merge this ahead of the other open PRs.

`SECURITY.md` directed vulnerability reports to `oscar@btq.com` and `barney@btq.com`. **`btq.com` is no longer ours, and it still accepts mail** — it carries live Google Workspace MX records:

```
1  aspmx.l.google.com.
5  alt1.aspmx.l.google.com.
5  alt2.aspmx.l.google.com.
10 aspmx2.googlemail.com.
10 aspmx3.googlemail.com.
```

So anyone following the project's own responsible-disclosure process has been sending vulnerability reports about BTQ to a third party. This needs no attacker action — it collects other researchers' findings passively.

`GOVERNANCE.md` named the same two addresses across six separate roles, and `WHITEPAPER_INTEGRATION_DESIGN.md` for maintainer contact.

## Changes

- **Reporting intake → `security@btq.tech`.** An alias rather than an individual, so the channel survives people changing roles — which is the failure mode that produced this.
- Named role holders in `GOVERNANCE.md` and the whitepaper doc move to their `btq.tech` addresses. `btq.tech` resolves (31.43.160.6) and has working Google Workspace MX.
- `GOVERNANCE.md`'s Security Officers section now states the intake alias alongside the officers who staff it, so the two documents agree on where a report actually goes. Previously `SECURITY.md` and `GOVERNANCE.md` could drift.

`security@btq.tech` needs to exist before or as this merges.

## Not fixed here

- **Mainnet and signet DNS seeds still point at `btq.com` subdomains** — #114. `seed1.btq.com` has no A record today, but adding one is the domain owner's choice, and it would hand them an eclipse position over every default-configured mainnet node. That fix needs seed infrastructure stood up, not just a text change, so it is deliberately separate.
- The tracked binaries under `build-linux/` (#108) contain `btq.com` strings, so shipped artifacts still carry the dead domain. Another reason to purge them.
- `SECURITY.md`'s **Supported Versions table is stale** — it lists 1.1.x, 1.0.x and 0.1.x with end-of-life dates of 2025-12-31 and 2024-12-31, both now past, against a repo whose tags run v0.1.0-testnet to v0.4.2-testnet. Left alone because I do not know the intended support policy, but it should not go in front of an external auditor as-is.

## Test plan

- [x] `grep -rn "btq\.com" doc-btq/ WHITEPAPER_INTEGRATION_DESIGN.md` returns nothing
[...]

Issue #145, opened 2026-08-05
supply chain: releases are unsigned, and verify.py defaults to domains we do not own


augchan42:
We document a binary verification process that has never been possible to complete.

Found while trying to publish the `security-keys.txt` that `doc-btq/SECURITY.md:81` promises. There are no keys to publish. Related: #118.

## What is missing

| | |
| --- | --- |
| `SHA256SUMS` in any release | **Never shipped.** v0.4.0, v0.3.2 and v0.3.0 carry `linux-x86_64.zip` and `windows-x86_64.zip` only |
| `SHA256SUMS.asc` | Never shipped |
| `contrib/builder-keys/` | Does not exist |
| Any PGP fingerprint in the tree | None |

`contrib/guix/` is vendored and includes `guix-attest`, so the deterministic-build machinery is present — it has just never produced a published attestation.

## The verify script defaults to domains we do not own

`contrib/verify-binaries/verify.py:49-50`:

```python
HOST1 = "https://btqcore.org"
HOST2 = "https://btq.org"
```

`btqcore.org` is **not registered by anyone** (WHOIS `Domain not found`, no NS, no A). `btq.org` was registered in 2000 and is not ours. Both are `bitcoin` → `btq` rename artifacts.

The hosts are baked into the tool, not just the docs. If anyone registers `btqcore.org`, they can serve a `SHA256SUMS` of their choosing to every user who runs our own verification script. Verification against attacker-supplied hashes is worse than none — it manufactures confidence.

Line 18 also points builder keys at `https://github.com/btq-core/guix.sigs/tree/main/builder-keys`. **That org 404s** — upstream is `bitcoin-core/guix.sigs`.

## Decisions needed first

1. **Do we sign releases?** If not, `verify.py` and the verification docs should be removed rather than left describing a process that does not exist.
2. **Who holds keys, and how many signers?** One key is a single point of failure and of compromise.
3. **Where are keys published?** `SECURITY.md` already promises `bitcoinquantum.com/security-keys.txt`.
4. **Guix attestations, or simple detached signatures?**

## Then

- [ ] Publish `SHA256SUMS` and `SHA256SUMS.asc` with releases
- [ ] Add `contrib/builder-keys/` with real fingerprints
- [ ] Repoint `HOST1` / `HOST2` to a host we control
- [ ] Fix or remove the `btq-core/guix.sigs` reference
- [ ] Publish `security-keys.txt` (unblocks #118)

## Audit relevance

[...]

Issue #118, opened 2026-07-31
supply chain: binary verification instructions fetch SHA256SUMS.asc from btq.org and btqcore.org, neither of which we own


augchan42:
Found while sweeping domain references for #115. Filed separately because this is a **release integrity** path, not a contact detail, and it needs a decision rather than a text substitution.

## What the docs tell users to do

```
contrib/README.md:39
  This script attempts to download and verify the signature file
  SHA256SUMS.asc from btq.org.

contrib/verify-binaries/README.md:26
  signature file `SHA256SUMS.asc` from https://btqcore.org and https://btq.org.
```

So our documented binary-verification procedure fetches the signature file — the artifact whose entire purpose is to establish that a release is genuine — from `btq.org` and `btqcore.org`.

## Why that is a problem

**We are not known to control either domain.**

Three domains are in play across the repository and only one is confirmed ours:

| domain | status |
| --- | --- |
| `btq.tech` | ours — resolves, working Google Workspace mail |
| `btq.com` | **lost** (confirmed) — still accepts mail, see #115 |
| `btq.org` | **unknown** — resolves to 3.33.251.168 / 15.197.225.128 (AWS ranges typical of parked domains), no MX records |

`btqcore.org` has not been checked.

If `btq.org` is not ours, then anyone who holds it can serve a `SHA256SUMS.asc` of their choosing to every user who follows our own verification instructions. Verification against attacker-supplied hashes is worse than no verification, because it produces false confidence.

This is inert only for as long as nobody actually runs the flow against a live host.

## Likely origin

These lines are almost certainly rename artifacts rather than deliberate choices. `contrib/verify-binaries` upstream references `bitcoincore.org` and `bitcoin.org`; a global `bitcoin` → `btq` substitution produces exactly `btqcore.org` and `btq.org`. The same rename left ~40 mangled URLs in `doc/release-notes/`, including `github.com/btq/btq.org/pull/279`, which was `bitcoin/bitcoin.org`.

That makes it an accident, not a decision — which is precisely why it has survived unexamined.

## What is needed

Not a find-and-replace. A decision:

- [ ] **Confirm registrar ownership of `btq.org` and `btqcore.org`.** Same check as `btq.com` in #115.
- [ ] Decide where releases and `SHA256SUMS.asc` are actually published. `bitcoinquantum.com` is the obvious candidate — it is where the website, docs, explorer and working testnet seed already live.
[...]

PR #148, opened 2026-08-05
doc: settle security contact on security@btq.tech, drop PGP

augchan42:
Two maintainer decisions, applied to every doc that carries a security contact.

## 1. Contact settles on `security@btq.tech`

**Changed after review — this PR originally moved contacts to `security@btq.com`.** btq.tech is the domain we use, and the repo already said so in two places:

- `MAINNET_AUDIT_MATRIX.md` records `btq.com` as a domain the project no longer controls, which directly contradicted this PR's original rationale.
- Every maintainer, release manager and security officer address in `doc-btq/GOVERNANCE.md` is already `@btq.tech`, leaving `security@btq.com` as the only address on a different domain.

So this keeps the contact where #115 put it on 31 July rather than moving it a second time. No mailbox needs provisioning, which was the open blocker on the original version.

Updated in `SECURITY.md`, `doc-btq/SECURITY.md`, `doc-btq/GOVERNANCE.md`, `WHITEPAPER_INTEGRATION_DESIGN.md`, `doc-btq-design.md`.

The `seed1.btq.com` references in `MAINNET_AUDIT_MATRIX.md` are deliberately untouched — they describe the broken mainnet DNS seed tracked in #114, not a contact address.

## 2. PGP is removed, not deferred

We are not doing PGP, so the docs should stop implying we might. Removed:

| Removed | Why |
| --- | --- |
| `## PGP Keys` → Security Officers block in `doc-btq/SECURITY.md` | Every field was `[TO BE ADDED]`, including the fingerprint. A placeholder fingerprint is worse than no section — it reads as a key that exists but wasn't pasted. |
| Key-verification steps pointing at `https://bitcoinquantum.com/security-keys.txt` | **Returns 404.** Instructions to `gpg --import` a file that isn't there. |
| `[PGP signatures from security officers]` in the advisory template | Advisories aren't signed. |
| `Signal: [TO BE ADDED] (for encrypted communication)` | Same placeholder problem. |
| "does not yet publish PGP keys … email the address above and ask" in `SECURITY.md` | Promised a capability we don't have. "Ask and we'll sort something out" costs the first exchange with every serious reporter. |

Replaced with a plain statement that encrypted intake is not offered, plus a route that costs nothing and preserves the option:

> If a report is sensitive enough that plain email is unacceptable, say so in a first message containing no detail, and we will agree a channel before you send anything substantive.

Release-signing references in `doc-btq-design.md` are untouched — build provenance is a separate question from disclosure intake, and this PR takes no position on it.

## Also worth knowing

Before #115, `SECURITY.md` pointed reports at `security@btqcore.org` and listed the PGP fingerprints of three upstream Bitcoin Core developers — Pieter Wuille, Michael Ford, Andrew Chow — inherited with the fork and never ours. Both are already gone from `master`; noting it because **`btqcore.org` is currently unregistered**, so that address survives in git history, forks, and old release tarballs. Anyone who registers the domain can receive mail sent to it. That is tracked separately and is not something this PR can fix.

## Verification

- `grep -rn "security@btq" --include="*.md"` — only `security@btq.tech` remains outside `src/secp256k1/` (upstream, deliberately untouched)
- No `[TO BE ADDED]` placeholders left in either SECURITY.md
- Docs only; no code paths touched

**Note:** the branch is still named `docs/security-contact-btq-com` from the original approach. Renaming it would break the existing review links, so it is left as-is.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -15,7 +15,7 @@ Two private channels, in order of preference:
    button under this repository's Security tab. Preferred: the report stays
    private, the thread is authenticated, and the advisory and CVE are issued
    from the same place.
-2. **Email security@btq.com** — for security reports only, not support.
+2. **Email security@btq.tech** — for security reports only, not support.
 
 `doc-btq/SECURITY.md` documents the severity categories, response timelines
 and coordinated-disclosure process.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -9,13 +9,13 @@ only the most recent release line receives security fixes.
 
 **Do not report security vulnerabilities through public GitHub issues.**
 
-Email **security@btq.tech**. This address is for security reports only, not
+Email **security@btq.com**. This address is for security reports only, not
 for support.
 
 `doc-btq/SECURITY.md` documents the severity categories, response timelines
 and coordinated-disclosure process.
 
-## PGP keys
+## Encryption
 
-BTQ Core does not yet publish PGP keys for its security contacts. If you need
-to encrypt a report before that changes, email the address above and ask.
+BTQ Core does not publish PGP keys and does not offer encrypted intake. Send
+reports in plain email to the address above.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -9,13 +9,19 @@ only the most recent release line receives security fixes.
 
 **Do not report security vulnerabilities through public GitHub issues.**
 
-Email **security@btq.com**. This address is for security reports only, not
-for support.
+Two private channels, in order of preference:
+
+1. **GitHub private vulnerability reporting** — the "Report a vulnerability"
+   button under this repository's Security tab. Preferred: the report stays
+   private, the thread is authenticated, and the advisory and CVE are issued
+   from the same place.
+2. **Email security@btq.com** — for security reports only, not support.
 
 `doc-btq/SECURITY.md` documents the severity categories, response timelines
 and coordinated-disclosure process.
 
 ## Encryption
 
-BTQ Core does not publish PGP keys and does not offer encrypted intake. Send
-reports in plain email to the address above.
+BTQ Core does not publish PGP keys. Use GitHub private reporting if you need
+the exchange to stay confidential — it is private end to end and requires no
+key management. Email to the address above is unencrypted.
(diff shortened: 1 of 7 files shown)


augchan42:
**Two merge prerequisites, both settings changes rather than code.**

**1. Enable private vulnerability reporting** (Settings → Security & analysis). Currently **off** — the docs now point at a "Report a vulnerability" button that will not exist until it is flipped on. Needs admin.

This is the replacement for the PGP intake this PR removes: private end to end, authenticated, no key management, and the advisory and CVE issue from the same thread. `doc-btq/SECURITY.md` already listed GitHub Security Advisories as a *publication* channel — this closes the intake half.

**2. Add a DMARC record to btq.com.** Checked 2026-08-05:

```
btq.com    SPF: yes    DMARC: none    MTA-STS: none
btq.tech   DMARC: p=quarantine
```

We are about to publish `security@btq.com` on a domain with no DMARC policy, so mail *from* that address can be spoofed. The concrete risk is a forged "thanks, known duplicate, feel free to disclose" sent to a reporter mid-embargo — an attack on the disclosure process itself. `p=none` with a `rua` first, then tighten, is enough to close it.

Tracked in BTQ-157 alongside provisioning the mailbox.

**On PGP vs DMARC**, since it came up: they are different layers. DMARC authenticates outbound mail claiming to be from our domain; it provides no confidentiality and does nothing for a report arriving from a researcher's own mailbox. The reason to drop PGP is that nobody uses it and a stale key is worse than none — not that DMARC covers it. GitHub PVR is what actually covers it.

BarneyChambers:
This moves the security contact in the opposite direction to the domain sweep. The sweep is `btq.com -> bitcoinquantum.com` / `btq.tech`, and `btq.tech` is the address we should be publishing, so this PR replaces the correct domain with the one we are migrating away from.

Suggest inverting it: keep `security@btq.tech` throughout, and keep whatever else is here that is independently worth doing.

Two other things in the diff that are separate decisions rather than a domain rename, worth splitting out or calling out explicitly so they are not merged implicitly:

1. Dropping PGP intake. Removing the security officers' keys means all vulnerability reports arrive unencrypted. That is a real policy choice for a project shipping post-quantum signatures, and reads oddly in that context.
2. `https://bitcoinquantum.com/security-keys.txt` is removed along with the keys. If PGP does stay, that URL is on the domain we are moving *to*, so it is the one part of the current text that already matches the sweep.

Related: #145 and #118 both cover published domains we do not own, and should probably be resolved in the same pass so we only touch these files once.

Issue #146, opened 2026-08-05
node: -blocksonly=1 makes btqd refuse to start on every network


BarneyChambers:
`btqd -blocksonly=1` exits at init on every chain, with no way to start short of also passing `-maxmempool` by hand. Found while trying to bring up a throwaway testnet node.

```
$ ./src/btqd -regtest -datadir=/tmp/x -blocksonly=1
InitParameterInteraction: parameter interaction: -blocksonly=1 -> setting -maxmempool=5
Error: -maxmempool must be at least 80 MB
Shutdown: In progress...
```

## Why

Two constants that have to agree no longer do, because only one of them was adjusted for BTQ's larger blocks.

`init.cpp:752` soft-sets the mempool ceiling when blocksonly is on:

```cpp
if (args.SoftSetArg("-maxmempool", ToString(DEFAULT_BLOCKSONLY_MAX_MEMPOOL_SIZE_MB)))
```

`DEFAULT_BLOCKSONLY_MAX_MEMPOOL_SIZE_MB` is still upstream's `5` (`kernel/mempool_options.h:21`).

`init.cpp:1490` then requires the mempool to hold 40 descendant packages:

```cpp
int64_t descendant_limit_bytes = mempool_opts.limits.descendant_size_vbytes * 40;
if (mempool_opts.max_size_bytes < 0 || mempool_opts.max_size_bytes < descendant_limit_bytes) {
    return InitError(strprintf(_("-maxmempool must be at least %d MB"), ...));
}
```

BTQ raised `DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` along with `MAX_BLOCK_WEIGHT`, so that floor is 80 MB here against upstream's ~4 MB. Upstream is fine because 5 > 4. We are not, because 5 < 80.

## Why it matters beyond the inconvenience

`-blocksonly` is what you run on a node that should not maintain a relay mempool — an explorer backend, a mining node's block source, a resource-constrained validator. It is also a standard first thing to reach for when bootstrapping, which is how I hit it. Right now the documented option cannot be used at all, and the error message points at `-maxmempool`, which the operator never set.

## Fix

Scale the blocksonly default the same way the descendant limit was scaled, rather than leaving one at upstream's value. Something like deriving it from the descendant floor instead of hardcoding, so the two cannot drift apart again:

```cpp
// Must clear the descendant-package floor enforced in AppInitMain, which BTQ
[...]

PR #149, opened 2026-08-05
CVE-2025-46598 parts 2 and 3: sighash midstate cache, single script run, no tx peer punishment

BarneyChambers:
Finishes CVE-2025-46598 (BTQ-140). Part 1, witness-stripping detection without re-running scripts (upstream #33105), landed in #139. This is the other two mitigations named in the [Bitcoin Core advisory](https://bitcoincore.org/en/2025/10/24/disclose-cve-2025-46598/).

The CVE is a CPU DoS: specially-crafted unconfirmed transactions take seconds each to validate, get rejected without costing the sender anything, and can be repeated to delay block propagation.

## Part 2 — sighash midstate cache (upstream #32473)

Legacy sighash re-serialises the entire transaction on every call, so a script forcing many sighashes over one input, via `OP_CODESEPARATOR` or simply many signatures, costs quadratic work. `SigHashCache` keeps the SHA256 state from just before the sighash type byte is appended, so a repeated script code within one input serialises the transaction once.

Measured on a 200-input transaction taking 20k sighashes over one script code: **147ms -> 2.4ms, 61x**.

**Divergence from upstream, deliberate.** Upstream keys the cache on sighash mode and script code only, on the grounds that no input can reach both the `BASE` and `WITNESS_V0` paths. That does not hold here: `CheckDilithiumSignature` dispatches on `sigversion` at runtime inside a single input's checker. We key on sig version too. Getting this wrong would validate a signature against a sighash computed for the other sig version, so `sighash_cache_keying` fails if the sig version is dropped from the key.

## Part 3 — one script run, no peer punishment (upstream #33050)

A failing input script was validated twice: once with policy flags, then again with them masked off, solely to decide whether to discourage the relaying peer. That doubles the CPU an attacker can make us spend on a transaction we reject either way, and the punishment bought little, since an attacker can place a non-standard input first and evade it entirely while an honest peer running slightly different relay policy gets disconnected. `MaybePunishNodeForTx` and the second run are both gone. Block-based discouragement (`MaybePunishNodeForBlock`) is untouched.

**Divergence from upstream, and this one is load-bearing.** Upstream picks the error vocabulary by testing `flags & STANDARD_NOT_MANDATORY_VERIFY_FLAGS`, treating a standard-not-mandatory flag as proof of mempool context. In BTQ `SCRIPT_VERIFY_DILITHIUM_P2MR_ONLY` is standard-but-not-mandatory and yet `GetBlockScriptFlags` sets it once `DEPLOYMENT_DILITHIUM_P2MR` activates, so a literal port made **block** validation report consensus failures as mempool ones. `feature_csv_activation` caught it. Fixed by passing the context explicitly, which cannot rot as flags move between the sets.

Worth noting separately: this also removes a trap the old code had documented in `policy.h`. The retry masked off `SCRIPT_VERIFY_DILITHIUM_P2MR_ONLY` even during block validation, so a block transaction failing only that consensus rule could be reported `TX_NOT_STANDARD`. With no retry, mempool failures are unconditionally `TX_NOT_STANDARD` and block failures unconditionally `TX_CONSENSUS`, and neither rests on flag placement any more.

Mempool script failures are now reported as `mempool-script-verify-flag-failed`, matching upstream.

## Testing

All 769 unit tests pass.

Three mutations of the sighash cache, each caught: ignoring sig version (8017 failing assertions), ignoring script code (1944), storing the midstate after the type byte (11639).

Functional tests this touches, before and after:

| Test | Before | After |
|---|---|---|
| `rpc_packages` | pass | pass |
| `feature_dersig` | pass | pass |
| `feature_csv_activation` | pass | pass |
| `feature_block` | pass | pass |
| `p2p_invalid_tx` | fail | **pass** |
| `p2p_segwit` | fail | fail |
| `feature_segwit` | fail | fail |
| `feature_cltv` | fail | fail |
| `feature_nulldummy` | fail | fail |
| `mempool_accept` | fail | fail |

The five still failing were already failing on master for the #104 subsidy mismatch and fail at exactly the same line as before. Their reject strings are updated for the rename, but I could not exercise those assertions, so they are the part of this PR to look at hardest.

`p2p_invalid_tx` went green because I fixed its #104 amounts to derive from `block_subsidy_sat` — needed to reach the peer-connection assertion this PR changes. Its orphan-peer id now comes from `getpeerinfo` rather than a hardcoded 25, which shifted once the reconnects went away.

## Review notes

- Consensus behaviour is unchanged: in block validation the branch taken and the resulting `TX_CONSENSUS` are the same as before, only the second script run is gone.
- The policy change is real and worth a deliberate ack: a peer relaying a consensus-invalid transaction is no longer disconnected. Upstream's reasoning is in #33050; the short version is that this protection was already evadable and cost us double CPU to enforce.

Closes BTQ-140.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -557,14 +557,6 @@ private:
                                  bool via_compact_block, const std::string& message = "")
         EXCLUSIVE_LOCKS_REQUIRED(!m_peer_mutex);
 
-    /**
-     * Potentially disconnect and discourage a node based on the contents of a TxValidationState object
-     *
-     * @return Returns true if the peer was punished (probably disconnected)
-     */
-    bool MaybePunishNodeForTx(NodeId nodeid, const TxValidationState& state)
-        EXCLUSIVE_LOCKS_REQUIRED(!m_peer_mutex);
-
     /** Maybe disconnect a peer and discourage future connections from its address.
      *
      * @param[in]   pnode     The node to check.
@@ -1799,32 +1791,6 @@ bool PeerManagerImpl::MaybePunishNodeForBlock(NodeId nodeid, const BlockValidati
     return false;
 }
 
-bool PeerManagerImpl::MaybePunishNodeForTx(NodeId nodeid, const TxValidationState& state)
-{
-    PeerRef peer{GetPeerRef(nodeid)};
-    switch (state.GetResult()) {
-    case TxValidationResult::TX_RESULT_UNSET:
-        break;
-    // The node is providing invalid data:
-    case TxValidationResult::TX_CONSENSUS:
-        if (peer) Misbehaving(*peer, 100, "");
-        return true;
-    // Conflicting (but not necessarily invalid) data or different policy:
-    case TxValidationResult::TX_RECENT_CONSENSUS_CHANGE:
-    case TxValidationResult::TX_INPUTS_NOT_STANDARD:
-    case TxValidationResult::TX_NOT_STANDARD:
-    case TxValidationResult::TX_MISSING_INPUTS:
-    case TxValidationResult::TX_PREMATURE_SPEND:
-    case TxValidationResult::TX_WITNESS_MUTATED:
-    case TxValidationResult::TX_WITNESS_STRIPPED:
-    case TxValidationResult::TX_CONFLICT:
-    case TxValidationResult::TX_MEMPOOL_POLICY:
-    case TxValidationResult::TX_NO_MEMPOOL:
-        break;
-    }
-    return false;
-}
-
 bool PeerManagerImpl::BlockRequestAllowed(const CBlockIndex* pindex)
 {
     AssertLockHeld(cs_main);
@@ -3032,8 +2998,6 @@ bool PeerManagerImpl::ProcessOrphanTx(Peer& peer)
                     orphan_wtxid.ToString(),
                     peer.m_id,
                     state.ToString());
-                // Maybe punish peer that gave us an invalid orphan tx
-                MaybePunishNodeForTx(peer.m_id, state);
             }
             // Has inputs but not accepted to mempool
             // Probably non-standard or insufficient fee
@@ -4365,7 +4329,6 @@ void PeerManagerImpl::ProcessMessage(CNode& pfrom, const std::string& msg_type, 
                 tx.GetWitnessHash().ToString(),
                 pfrom.GetId(),
                 state.ToString());
-            MaybePunishNodeForTx(pfrom.GetId(), state);
         }
         return;
     }
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -1783,10 +1783,46 @@ bool SignatureHashSchnorr(uint256& hash_out, ScriptExecutionData& execdata, cons
     return true;
 }
 
+int SigHashCache::CacheIndex(int hash_type) noexcept
+{
+    return 3 * !!(hash_type & SIGHASH_ANYONECANPAY) +
+           2 * ((hash_type & 0x1f) == SIGHASH_SINGLE) +
+           1 * ((hash_type & 0x1f) == SIGHASH_NONE);
+}
+
+bool SigHashCache::Load(int hash_type, SigVersion sigversion, const CScript& script_code, HashWriter& writer) const
+{
+    const auto& entry = m_cache_entries[CacheIndex(hash_type)];
+    if (!entry.has_value()) return false;
+    if (entry->sigversion != sigversion) return false;
+    if (entry->script_code != script_code) return false;
+    writer = HashWriter(entry->writer);
+    return true;
+}
+
+void SigHashCache::Store(int hash_type, SigVersion sigversion, const CScript& script_code, const HashWriter& writer)
+{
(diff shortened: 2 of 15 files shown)


BarneyChambers:
Validation against current master (`d5d1b1e8`):

| Case | Result |
|---|---|
| #149 alone | merges clean, builds, 769 unit tests pass |
| #149 + #143 | merges clean, builds, 769 unit tests pass |

The combined case is checked deliberately: two individually-green PRs broke master earlier today through a semantic conflict that neither one's CI could see, so "both green separately" is not sufficient evidence on this repo.

Functional, on the combined tree: `rpc_packages`, `feature_dersig`, `feature_csv_activation`, `feature_block`, `p2p_invalid_tx` all pass. `mining_basic.py` fails identically on plain master (`test_blockmintxfee_parameter`, line 115), so it is pre-existing.

Mutation testing on the sighash cache, since a silently-wrong cache would be a consensus bug rather than a test failure. Each mutation must be caught, and each is:

| Mutation | Result |
|---|---|
| ignore sig version when matching | caught, 8017 failing assertions |
| ignore script code when matching | caught, 1944 failing assertions |
| store midstate after the type byte | caught, 11639 failing assertions |

The first of those is the one that matters for the upstream divergence: it is the check upstream does not need and we do.

BarneyChambers:
The red `test each commit` here is not this PR. It fails in `sigopcount_tests/dilithium_witness_v0_sigop_weighting` with `std::length_error`, in a file this PR does not touch.

That test builds a witness element from `ecdsa_key.GetPubKey().begin()` and `ecdsa_key.GetPubKey().end()`. `GetPubKey()` returns by value, so those are two separate temporaries and the distance between them is whatever the stack layout gives. The job builds with clang and gets a bad layout; our gcc builds get a working one, which is why it passes locally and in the address-tests job.

Fix is in #151, with the mechanism reproduced. Same failure has been marking unrelated PRs red, including docs-only #148. This PR should go green once #151 lands.

BarneyChambers:
Rebased onto master (`130c44288`) now that #151 has landed. That cleared the
`sigopcount_tests` failure this PR was previously red on, and `test each
commit` gets considerably further. It still fails, but no longer on anything
from this branch.

The remaining failure is `make check` aborting in the benchmark sanity check:

```
bench_btq: bench/checkblock.cpp:50: Assertion `checked' failed.
make[3]: *** [Makefile:23572: check-local] Aborted (core dumped)
```

That reproduces on master by itself, with none of this branch's commits:

```
$ git checkout origin/master     # 130c44288
$ ./src/bench/bench_btq -sanity-check -priority-level=high
Error: CreateNewBlock: TestBlockValidity failed: time-too-new, block timestamp too far in the future
$ echo $?
1
```

So `test each commit` cannot currently pass for any PR. Filed as #154 with the
diagnosis: these are upstream benchmarks that were never adapted to BTQ, and
`DeserializeAndCheckBlockTest` in particular asserts that an upstream **Bitcoin**
block passes **BTQ** consensus. It was hidden behind the sigopcount failure
until #151 merged.

For this PR specifically, the per-commit checks were run locally instead.
Both commits, rebased and taken one at a time, build clean and pass
`sigopcount_tests`, `sighash_tests` and `txvalidationcache_tests`.
`Build btqd and run address tests` is green here.

BarneyChambers:
The bench breakage blocking this is fixed in #155 (issue #154). Once that merges I will rebase this onto master and re-run CI.

For the record, the `test each commit` failure here was never caused by this PR: `make check` runs `bench_btq -sanity-check`, and two upstream benchmark fixtures abort under BTQ's consensus rules. A 1 MB Bitcoin block weighs 15,998,192 against BTQ's 8,000,000 cap because `WITNESS_SCALE_FACTOR` is 16, and the wallet benchmark mines to a `bcrt1...` address that BTQ regtest cannot decode. Both were masked until the sigopcount fix in #151 let `make check` get that far.

BarneyChambers:
Rebased onto master now that #155 has landed. `make check` passes, so the benchmark abort is gone. The `test each commit` job is still red, but for a reason that has nothing to do with this PR: it now reaches the functional suite, which is broken repo-wide under #104.

I ran the full functional suite locally against `origin/master` (0604161d4) and against this branch, same machine and same build settings, and diffed the failure sets:

| | failures |
|---|---|
| master | 102 |
| this branch | 100 |
| regressions introduced here | 0 |
| tests this branch fixes | 2 (`p2p_invalid_tx.py`, `p2p_invalid_tx.py --v2transport`) |

The 100 shared failures are the #104 set, dominated by the 5-vs-50 subsidy mismatch. The CI log shows the same signatures: `AssertionError: not(5.00000000 == 50)`, `-maxmempool must be at least 80 MB` (#146), and the RPC client conversion table mismatch from `importdilithiumkey`'s missing `rescan` entry.

Worth flagging separately: `test each commit` is the only job that runs the functional suite, and it skips on single-commit PRs. #143, #151 and #153 all skipped it, which is why this is the first PR to surface the breakage. Until #104 is closed, that job cannot pass for any multi-commit PR.

BarneyChambers:
Merging with `test each commit` red. The evidence above is the basis: zero regressions against master, two tests fixed, and the remaining failures are the pre-existing #104 set that no multi-commit PR can currently clear. `make check` and the unit suite both pass.

PR #163, opened 2026-08-13
psbt: carry Dilithium P2MR signing material across wallets

### LINKED ISSUE #85: psbt: P2MR spend data cannot cross devices — no PSBT fields for merkle path, leaf script or Dilithium key material

Oskii:
## Summary
- A PSBT can now carry the P2MR leaf, control block and Dilithium partial signatures between wallets, so a 2-of-3 no longer needs the `.btqms` workaround. Fail-closed parse: a forged signature is rejected when the PSBT is read, not when it is broadcast.
- `createdilithiummultisig` / `getdilithiumpubkey` plus the existing `walletcreatefundedpsbt` → `walletprocesspsbt` → `combinepsbt` → `finalizepsbt` flow are the RPCs. `doc/psbt.md` has the copy-paste `btq-cli` recipe; `contrib/btq/dilithium-psbt.sh` wraps it so the ~16 KB payload stays in a file.
- Parse cap on control blocks is the consensus limit (`P2MR_CONTROL_MAX_SIZE`), not 513 bytes, so a deep leaf that is spendable on-chain is also spendable through a PSBT.

Fixes #85.

## Test plan
- [ ] `src/test/test_btq --run_test=psbt_dilithium_tests`
- [ ] `test/functional/wallet_dilithium_psbt.py`
- [ ] `test/functional/wallet_dilithium_psbt_multisig.py`
- [ ] `test/functional/tool_dilithium_psbt.py`
- [ ] Follow `doc/psbt.md` “Dilithium P2MR” against a regtest node using only `btq-cli`
- [ ] `decodepsbt` on a 1-of-2 signed PSBT shows `status: partially_signed` and `signatures: 1`; a flipped byte in the signature is rejected at decode

### DIFF
--- a/contrib/btq/dilithium-psbt.sh
+++ b/contrib/btq/dilithium-psbt.sh
@@ -1,0 +1,363 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026 The BTQ Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#
+# Operator wrapper around the Dilithium P2MR PSBT RPCs. Every command is a
+# thin call to btq-cli; nothing here signs except walletprocesspsbt.
+#
+# A 2-of-3 PSBT is ~16 KB. Keep it in a file, not a QR code.
+#
+# Usage:
+#   dilithium-psbt.sh --cli="<btq-cli invocation>" <command> [args]
+#   e.g. dilithium-psbt.sh --cli="src/btq-cli -testnet" pubkey alice
+#
+# Commands print JSON or a PSBT/hex payload on stdout. Diagnostics go to stderr.
+# See doc/psbt.md ("Dilithium P2MR") for the equivalent raw btq-cli recipe.
+
+set -euo pipefail
+
+usage() {
+    cat <<'EOF'
+usage: dilithium-psbt.sh --cli="<btq-cli invocation>" <command> [args]
+
+Commands (each is one btq-cli RPC, or a short sequence of them):
+
+  pubkey   <wallet> [label]
+      New Dilithium key in <wallet>. Prints address, p2mr_id, pubkey.
+      RPCs: getnewdilithiumaddress, getdilithiumpubkey
+
+  register <wallet> <m> <pubkey> [<pubkey>...]
+      Register the same m-of-n P2MR destination in <wallet>.
+      Every co-signer must pass the same m and the same pubkeys in the same order.
+      RPC: createdilithiummultisig
+
+  create   <wallet> <txid:vout> <address> <amount>
+      Unsigned funded PSBT spending that outpoint to <address>.
+      RPC: walletcreatefundedpsbt
+
+  sign     <wallet> <psbt>
+      Add this wallet's Dilithium signature(s). <psbt> is a file or base64.
+      RPC: walletprocesspsbt
+
+  combine  <psbt> <psbt> [<psbt>...]
+      Merge independently signed copies of the same unsigned PSBT.
+      RPC: combinepsbt
+
+  status   <psbt>
+      Per-input Dilithium signing progress (use this, not analyzepsbt).
+      RPC: decodepsbt
+
+  finalize <psbt>
+      Build the witness. Prints the raw hex when complete.
+      RPC: finalizepsbt
+
+  broadcast <hex>
+      RPC: sendrawtransaction
+
+Pass a PSBT as a file path (preferred) or as a base64 string.
+EOF
+}
+
+CLI=()
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -h|--help)
+            usage
+            exit 0
+            ;;
+        --cli=*)
+            # shellcheck disable=SC2206
+            CLI=(${1#--cli=})
+            shift
+            ;;
+        --cli)
+            # shellcheck disable=SC2206
+            CLI=($2)
+            shift 2
+            ;;
+        --)
+            shift
+            break
+            ;;
+        -*)
+            echo "unknown option: $1 (btq-cli flags belong inside --cli)" >&2
+            exit 1
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
+
+if [[ ${#CLI[@]} -eq 0 ]]; then
+    if [[ -n "${BTQ_CLI:-}" ]]; then
+        # shellcheck disable=SC2206
+        CLI=($BTQ_CLI)
+    else
+        CLI=(btq-cli)
+    fi
+fi
+
+if [[ $# -lt 1 ]]; then
+    usage >&2
+    exit 1
+fi
+
+need_python() {
+    command -v python3 >/dev/null 2>&1 || {
+        echo "python3 is required to parse RPC JSON" >&2
+        exit 1
+    }
+}
+
+# If btq-cli printed a JSON value, unwrap a top-level string so stdout is
+# the raw PSBT / txid / hex rather than a quoted JSON string.
+unwrap() {
+    need_python
+    python3 -c '
+import json, sys
+raw = sys.stdin.read()
+try:
+    value = json.loads(raw)
+except json.JSONDecodeError:
+    sys.stdout.write(raw)
+    raise SystemExit(0)
+if isinstance(value, str):
+    sys.stdout.write(value)
+    if not value.endswith("\n"):
+        sys.stdout.write("\n")
+else:
+    json.dump(value, sys.stdout)
+    sys.stdout.write("\n")
+'
+}
+
+# Read a dotted path out of JSON on stdin. Numeric path segments index arrays.
+jget() {
+    need_python
+    python3 -c '
+import json, sys
+data = json.load(sys.stdin)
+cur = data
+for part in sys.argv[1].split("."):
+    if part == "":
+        continue
+    if isinstance(cur, list):
+        cur = cur[int(part)]
+    else:
+        cur = cur[part]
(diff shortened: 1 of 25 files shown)


augchan42:
Approach ACK e78994462cd5a1a0336a14cd27e603bf8e8aae86

The approach is sound and the tests are thorough. One defect below needs a fix before I can ACK.

I ran an automated review pass, then checked each item against the code and ran the tests. Only the items below come from this branch; I have left the rest out.

## One real defect: a leaf with no control block

`ValidateP2MRDilithiumInput` checks that a leaf commits to the witness program, but the check sits inside the per-control-block loop (`src/psbt_dilithium.cpp:158-176`):

```cpp
for (const auto& [leaf, control_blocks] : input.m_p2mr_scripts) {
    const auto& [script, leaf_ver] = leaf;
    const uint256 leaf_hash = ComputeTapleafHash(static_cast<uint8_t>(leaf_ver), script);
    for (const auto& control : control_blocks) {
        ...
        if (ComputeP2MRMerkleRoot(control, leaf_hash) != uint256(program)) {
            return fail("P2MR leaf script does not commit to the witness program");
        }
    }
    if (!leaves.emplace(leaf_hash, CScript(script.begin(), script.end())).second) {
```

If `control_blocks` is empty, the inner loop never runs. The leaf still goes into `leaves`, so it counts as checked, and a partial signature can then refer to it. `SignP2MR` later reads `*control_blocks.begin()` on that same empty set (`src/script/sign.cpp:531`), which is undefined behaviour.

The wire cannot carry this, because a leaf with no control block serializes no record. `PSBTInput::Merge` and in-process construction can still produce it, so the invariant is still worth holding.

Suggested fix. Reject the empty set in the validator, and skip such a leaf in the signer:

```diff
--- a/src/psbt_dilithium.cpp
+++ b/src/psbt_dilithium.cpp
@@ -157,6 +157,12 @@ bool ValidateP2MRDilithiumInput(const PartiallySignedTransaction& psbt, unsigned
     std::map<uint256, CScript> leaves;
     for (const auto& [leaf, control_blocks] : input.m_p2mr_scripts) {
         const auto& [script, leaf_ver] = leaf;
+        // A leaf with no control block proves nothing, because the commitment
+        // check below runs per control block and would not run at all. Reject
+        // it here so the leaf never reaches the signer as an accepted leaf.
+        if (control_blocks.empty()) {
+            return fail("P2MR leaf script has no control block");
+        }
         const uint256 leaf_hash = ComputeTapleafHash(static_cast<uint8_t>(leaf_ver), script);
         for (const auto& control : control_blocks) {
```

```diff
--- a/src/script/sign.cpp
+++ b/src/script/sign.cpp
@@ -525,6 +525,10 @@ static bool SignP2MR(const SigningProvider& provider, const BaseSignatureCreator
     std::vector<std::vector<unsigned char>> smallest_result_stack;
     for (const auto& [key, control_blocks] : sigdata.p2mr_spenddata.scripts) {
         const auto& [script, leaf_ver] = key;
+        // Spend data merged from a PSBT can carry a leaf with no control block.
+        // Such a leaf cannot produce a witness, so skip it rather than
+        // dereference an empty set below.
+        if (control_blocks.empty()) continue;
         std::vector<std::vector<unsigned char>> result_stack;
```

A regression test that covers both halves. The signing half needs a provider that holds the key but not the tree, because a provider with the tree fills the empty set in through `GetP2MRSpendData`:

```cpp
BOOST_AUTO_TEST_CASE(a_leaf_with_no_control_block_is_rejected)
{
    const Signer signer = MakeSigner();
    CScript leaf;
    leaf << ToByteVector(signer.pubkey) << OP_CHECKSIGDILITHIUM;
    Fixture f = MakeFixture(leaf, {signer});

    // No control block means nothing proves the leaf belongs to the output. The
    // wire cannot carry this, because a leaf with no control block serializes no
    // record, but a merge or an in-process build can still produce it.
    f.psbt.inputs[0].m_p2mr_scripts[{std::vector<unsigned char>(leaf.begin(), leaf.end()), TAPROOT_LEAF_TAPSCRIPT}] = {};

    const PrecomputedTransactionData txdata = PrecomputePSBTData(f.psbt);
    std::string error;
    BOOST_CHECK(!ValidateP2MRDilithiumInput(f.psbt, 0, &txdata, error));
    BOOST_CHECK_MESSAGE(error.find("no control block") != std::string::npos, error);

    // A signer that holds the key but not the tree takes the leaf straight from
    // the PSBT, so nothing fills the empty set in. Signing must decline the leaf
    // instead of reading the first control block of an empty set.
    FlatSigningProvider keys_only;
    const DilithiumPKHash id(signer.pubkey);
    keys_only.dilithium_pubkeys.emplace(id, signer.pubkey);
    keys_only.dilithium_keys.emplace(id, signer.key);
    BOOST_CHECK(!SignPSBTInput(keys_only, f.psbt, 0, &txdata, SIGHASH_ALL, nullptr, /*finalize=*/true));
    BOOST_CHECK(f.psbt.inputs[0].final_script_witness.IsNull());
}
```

I checked that the test earns its place: it fails with both guards removed and passes with them in.

## The design document now exists twice

This branch adds `doc-btq/DILITHIUM_PSBT_DESIGN.md`, and `design/DILITHIUM_PSBT_DESIGN.md` stays in the tree. Both carry the same title, and they disagree on normative points:

| Point | `design/` (Draft v3, 2026-07-30) | `doc-btq/` (v1, 2026-08-06) |
|---|---|---|
| Control block cap | 513 bytes | 4097 bytes, the consensus limit |
| `SIGHASH_DEFAULT` | Accepted, normalized to `0x01` | Rejected at parse |

The `doc-btq` copy is plainly the successor: its §16 lists these departures on purpose. Nothing in the tree links to the `design/` path, and the new `doc/psbt.md` points at the `doc-btq` copy. Two normative documents with no stated winner invite an implementation against the wrong one, so I suggest deleting `design/DILITHIUM_PSBT_DESIGN.md` in this PR. Git history keeps the draft.

## A contradiction inside the new document

`doc-btq/DILITHIUM_PSBT_DESIGN.md:226` (§2.5.5) still carries the draft's rule:

> PSBT parsers MAY accept `0x00` at parse time (normalized for sighash verification) but SHOULD warn; finalizer MUST rewrite to `0x01`.

The verification pseudocode just above it normalizes the same way. Both disagree with §4.6 of the same document, and with the code: `src/psbt_dilithium.cpp:202` rejects any sighash byte that is not `SIGHASH_ALL`. Suggested wording:

```
sig_raw   = partial_sig_value[0:2420]
hashtype  = partial_sig_value[2420]
REJECT if hashtype != 0x01
sighash   = SignatureHashSchnorr(..., hashtype, ...)
ACCEPT iff pubkey.Verify(sighash, sig_raw) == true
```

> **On-chain note:** Consensus `CheckDilithiumSignature` for `P2MR_TAPSCRIPT` **rejects** `SIGHASH_DEFAULT` (`0x00`) in the witness sighash byte, so a partial signature carrying `0x00` could never be spent. Parsers therefore reject `0x00` when the PSBT is read, as §4.6 states, rather than normalizing it to `0x01` and hiding the mistake until broadcast.

## Test results

I built the branch with the two guards and the new test, and ran:

- `psbt_dilithium_tests`: 13 of 13 pass
- All Dilithium, PSBT and signing suites: 104 of 104 pass
- `psbt_tests`, `script_tests`, `transaction_tests`, `sighash_tests`, `multisig_tests`: 41 of 41 pass
- `wallet_dilithium_psbt.py`, `wallet_dilithium_psbt_multisig.py`, `tool_dilithium_psbt.py`: all pass against a rebuilt `btqd`

Nice work on the test coverage here. The forged-signature and leaf-commitment cases made the empty-set gap easy to see.

## Out of scope

The automated pass also raised repository-wide items that this branch does not touch: no `AGENTS.md`, version claims in `doc-btq/README.md`, conflicting `SECURITY.md` files, broken links, and large compilation units. They are better as separate issues than as additions here.

PR #164, opened 2026-08-14
wallet: remove the unbuilt DilithiumWalletManager

augchan42:
## Summary

`src/wallet/dilithium_wallet_manager.{h,cpp}` are in no build target. Removing them, 300 lines.

## Evidence

```
$ grep -c dilithium_wallet_manager src/Makefile.am
0
$ grep -rl dilithium_wallet_manager $(git ls-files '*CMakeLists.txt')
(no output)
$ git grep -l 'dilithium_wallet_manager.h' -- src/ | grep -v dilithium_wallet_manager
(no output)
$ git grep -l DilithiumWalletManager -- src/
src/wallet/dilithium_wallet_manager.cpp
src/wallet/dilithium_wallet_manager.h
```

The class is named in no other file in the tree, nothing includes the header, and neither Autotools nor CMake compiles it. The code has never run.

## Why remove rather than wire up

Unbuilt code that looks live is worse than no code. These files read as the implementation of Dilithium wallet key handling, so anyone reasoning about that behaviour from them reaches conclusions the node does not implement. That has already happened during review here more than once.

The live path is `DescriptorScriptPubKeyMan` and `LegacyScriptPubKeyMan` in `src/wallet/scriptpubkeyman.cpp`, with the loaders in `src/wallet/walletdb.cpp`.

If the manager is wanted later, it should come back attached to a build target and a test, rather than sit in the tree looking finished.

## Risk

None. Nothing compiles or links these files, so the deletion cannot change behaviour. A build before and after produces the same objects.


BarneyChambers:
## Audit review

Verified the claims against `master` rather than reading the description. They hold. The file also could not have been built even if it were in a build target, which the description does not mention.

### What I checked

- The diff is two deletions and nothing else. `git diff --raw origin/master...HEAD` shows `D` for both files, 0 additions.
- `dilithium_wallet_manager` appears in neither `src/Makefile.am` (not in `BTQ_CORE_H`, not in `libbtq_wallet_a_SOURCES`) nor in any `*.include` or `CMakeLists.txt`.
- The build has no glob-based source lists that could pick it up implicitly. The only `$(wildcard ...)` uses in the tree are secp256k1 deps and Qt animation resources.
- `DilithiumWalletManager` is named in no other file, and nothing includes the header.

### The file does not compile

I ran the exact flags the build uses for `wallet/scriptpubkeyman.cpp`, with `-fsyntax-only`, against the file as it stands on `master`. Three hard errors:

- `LOCK(m_dilithium_mutex)` on a raw `std::mutex`. `UniqueLock` in `src/sync.h:152` derives from `MutexType::unique_lock`, which `std::mutex` does not have. Every method in the file uses this.
- `LegacyDilithiumBase58PaymentsAllowed` and `DilithiumLegacyDisabledError` are not declared in scope in `GetNewDilithiumAddress`.
- `DilithiumLegacyKeyID legacy_id(uint160(*dilithium_pkhash));` is a vexing parse, so it declares a function and the call on the next line cannot convert it.

Even if it compiled, `GetDilithiumKeyForDestination` and `GetDilithiumKeyForScript` take the lock and then call `GetDilithiumKeyLegacy`, which locks the same non-recursive mutex again.

This settles the "remove rather than wire up" question. 55a2ee5 ("wallet: keep Dilithium sends quantum-safe, and stop issuing dead addresses") added the P2MR-only guard and the dilithium-bech32 refusal to this file, referencing two functions that are not in scope there. A security fix was written into code that has never been compiled, so anyone who read that commit as covering the manager path was wrong.

### Fixed on the branch

`src/wallet/test/spend_tests.cpp` named the deleted file in a comment, which the first commit left dangling. Pushed 50a151e to fix it.

The count in that comment was wrong before the deletion too. Since the manager never compiled, the dilithium-bech32 refusal only ever existed in `LegacyScriptPubKeyMan::GetNewDestination` and `DescriptorScriptPubKeyMan::GetNewDestination`. That case also builds a descriptor-only wallet via `CreateDescriptorOnlyWallet`, so it exercises the descriptor manager, not both kinds. The new comment says that.

### Follow-ups, not blockers

The same problem sits one directory over and this PR does not touch it. `src/script/dilithium_signing_provider.{h,cpp}` and `src/crypto/dilithium_key_id.{h,cpp}` are also in no build target. `dilithium_signing_provider.h` is included only by its own `.cpp`, `DilithiumSigningProvider` is named nowhere else, and `DilithiumKeyID` / `DilithiumLegacyKeyID` are used only by those headers plus the manager being deleted here. The `DilithiumKeyID` hits in `p2mr.cpp`, `scriptpubkeyman.cpp` and `wallet.cpp` are substring matches inside names like `GetSingleDilithiumKeyIDForP2MR`, not the type. So this is a six-file dead cluster and the PR removes two. Worth a follow-up.

One inaccuracy left alone because it predates this branch: the dilithium-legacy comment just above says the refusal is asserted "in both spk manager kinds", but that block also runs on the descriptor-only wallet.

Nothing in `test/lint/` checks that a `src/*.cpp` belongs to a build target, which is how this survived. A lint for it would be cheap, though it will fail on the four files above until they are gone.

Approach and reasoning in the description are right. Good to merge once CI is green on the second commit.

BarneyChambers:
Note on the red check, since this merged with `test each commit` failing.

That job did not run on the first push, because it only runs once a PR has more than one commit. My comment fix gave it a second commit, so it started running and failed. The failure predates this branch.

Five functional tests failed: `feature_fee_estimation.py`, `feature_maxuploadtarget.py`, `feature_p2mr.py`, `mining_getblocktemplate_longpoll.py`, `p2p_segwit.py`. #167 is documentation-only and fails the same five. #163 fails the job as well. Tracked in #104.

Compilation was clean on both commits here, including `wallet/test/test_test_btq-spend_tests.o`, so the comment change builds. The deleted files were in no build target, so they cannot reach a functional test either way.

Worth noting for later: the `CI` workflow is skipped on pushes to master, so these five only surface on a PR that happens to carry two or more commits. A single-commit PR merges without anyone seeing them.

PR #167, opened 2026-08-18
doc: add protocol notes for the Dilithium rules

augchan42:
## What this adds

Two reference notes under `doc/protocol/`, kept next to the code so they stay true as it moves.

**`protocol-map.md`** maps each consensus rule to the area of the tree that defines it: signature
sizes, block and weight caps, sigops costs, address namespaces, P2MR, the opcode range, the two
activation deployments, key derivation, and the RPC surface. A reviewer can check a stated rule
against the code without grepping for it first.

**`pre-p2mr-script-templates.md`** records the Dilithium script templates that the P2MR-only rule
withdrew: P2DPK, Dilithium public-key-hash, witness-script spends, and P2DWPKH. It also records
why the rule is exclusivity and not merely P2MR, since BIP-360 does not cover that question.

## Why

The protocol description was carrying this material in prose, where it goes stale the first time
someone refactors. A path in a document that is versioned separately from the code is wrong as
soon as the code moves and nobody notices. Next to the code, a drift is visible in the diff.

The withdrawn templates are the part that earns its keep soonest. The public testnet contains
outputs of every type listed, so a node that syncs that chain has to validate them, and right now
that knowledge lives in people's heads and in the commit history.

## Scope

Documentation only. No code, no build, no test changes.

These are notes, not a specification. Where a note and the code disagree, the code is right and
the note is a bug. `protocol-map.md` says so, and asks that a change moving one of the named
areas update the row in the same pull request.

## Review

The rows in `protocol-map.md` are the part worth checking. Each names a rule, a value, and one or
more paths; the useful review is confirming the paths still hold the rule.


BarneyChambers:
Audited every row against the code at the base commit. The material is accurate: ML-DSA-44 sizes (1312/2560/2420), the 8 MB / 8 MW / scale-16 caps, 80,000 block sigops with `DILITHIUM_SIGOP_COST` 50, the 500-vs-50 tapscript validation weight and the 50-byte offset, 60-second spacing, 5 BTQ subsidy, the 2,100,000 halving interval, 21 M cap, 100-block maturity, LWMA window 144, testnet LWMA at 300,000, the 15,000-byte element limit, the `0xbb`-`0xbf` block, witness v2 with a 32-byte program, `SCRIPT_VERIFY_P2MR` set unconditionally, hardened-only `m/0'/0'/n'` and `m/0'/1'/n'`, and no `dumpdilithiumkey` all check out. The witness-v0 size-heuristic account in `pre-p2mr-script-templates.md` matches `interpreter.cpp:2202-2220`, including the pre-fork sigop weighting that `WitnessSigOps` preserves.

Six things did not, so I pushed three commits rather than leave the notes half true.

**The RPC row was the worst of it.** The seven P2MR calls live in `src/wallet/rpc/p2mr.cpp`, not `dilithium.cpp`. The row also named the deprecated `verifydilithiumsignature` and omitted `verifymessagewithdilithium` that replaced it in 3cd94f00f, and listed `signtransactionwithdilithium` flat when it is a stub that throws `RPC_METHOD_DEPRECATED` unconditionally (`dilithium.cpp:422`). Anyone using that row as a checklist would have got three wrong answers.

**"At genesis" is height 1.** `nDilithiumHeight`, `nDilithiumP2MRHeight` and `nLWMAHeight` are all `1` (`chainparams.cpp:98-101`), and the comment above them already says "LWMA from block 1". Harmless in effect, but a reader who greps has to reconcile, which is the friction the map exists to remove.

**The activation row omitted the testnet state, which is the one that matters.** On testnet `nDilithiumP2MRHeight` defaults to `INT_MAX` and is only set by `-testnetdilithiump2mrheight` (`chainparams.cpp:219`, `src/chainparams.cpp:26`). P2MR-only is unscheduled on the live public chain, and `pre-p2mr-script-templates.md` leans on that fact without the map recording it.

**"100,000-byte maximum script" is wrong for P2MR.** `MAX_SCRIPT_SIZE` and `MAX_OPS_PER_SCRIPT` are both gated on sigversion in `EvalScript` (`interpreter.cpp:481`, `:503`), so a P2MR leaf is subject to neither, except that `OP_CHECKMULTISIGDILITHIUM` still charges its key count against the 201 (`:1338`). Stated unqualified it was wrong in the one regime where Dilithium opcodes are valid.

**Two citations could not be resolved from the tree,** so I dropped them rather than guess. The Provenance section cited a transaction field guide dated 2026-07-28 that is not in the repo and has no link, and the reason it recorded is already ground 2 of the exclusivity argument. The weak-address argument was attributed twice under two names, "Aggarwal-Babbush classification" and "Babbush et al. §III.A", with no title or year either time. The property holds without either. Put them back with a real reference if you have the papers.

Smaller: "policy limit definitions" became `src/policy/policy.h`; `doc/README.md` now indexes `protocol/` next to `design/` and `policy/`; "BIP-360" to "BIP360" to match the rest of the tree; and the "recommended coinbase payout target" line now states the property instead, since that recommendation appears nowhere else in the repo.

Two things left, both better as separate changes:

1. Root-level `WHITEPAPER_INTEGRATION_DESIGN.md` still describes P2DPK and P2DWPKH as live output types. That is presumably the prose these notes replace, but nothing removes or cross-links it, so the tree now carries two descriptions and one is wrong.
2. `test/lint/lint-btq-consensus-constants.py` already pins `MAX_BLOCK_*`, `WITNESS_SCALE_FACTOR`, `COINBASE_MATURITY`, `MAX_SCRIPT_ELEMENT_SIZE`, `DILITHIUM_SIGOP_COST` and the initial subsidy against their C++ definitions. Teaching it to read the map's table would turn "Keeping it true" from a request into a check, and it already covers most of the numeric rows.

On CI: the "test each commit" failure is not from this branch. It is the functional suite (`feature_fee_estimation`, `feature_p2mr`, `mining_getblocktemplate_longpoll`, `wallet_txn_doublespend`, most dying in a second or two), and the docs-only `doc/purge-unowned-domains` branch failed the same check on Aug 11. Markdown cannot reach it.

Oskii (approved):
Documentation-only. Protocol-map rows were checked against current master (Dilithium sizes, 8 MB / 8 MW / scale 16, 5 BTQ subsidy, LWMA window 144, script limits, opcodes 0xbb-0xbf, HD paths, listed RPCs). The pre-P2MR note correctly records that public testnet still validates those output types.

`test each commit` is red on pre-existing functional tests, not on this diff. The lean build-and-address gate passed.

Follow-ups, not blockers: the RPC row does not list `verifymessagewithdilithium`; the activation row could repeat the testnet P2MR-only exception that the companion note already states.

Issue #168, opened 2026-08-18
Observation: heavy-work minority fork does not propagate; possible relevance to mainnet genesis


unkn0wn747:
## Background

Some background on how this branch arose. Starting from block 300,000 I was testing LWMA behaviour on testnet. During this, the block explorer stopped at 299,999 while I continued mining to block 300,360, reaching difficulty ~7,800 at block 300,360 over roughly 4 hours. I reported in Telegram that the explorer was stuck at 299,999 while my node was already at 300,360. The explorer then remained frozen at 300,067 for several days, while on my branch difficulty was declining back toward ~500 over the following days.

Now I see at the explorer that the height is 301 100 which is much higher than mine, while my chain still continues at my current height 300 447, but all the peers were lost. So it looks like we have a chain split (by some reasons, which I do not exactly know right now).

## Chain split

The accumulated work on the two branches is very different, which is what makes this case clear-cut. Chainwork is the sum of per-block difficulty, so a branch's weight is dominated by its high-difficulty blocks, not its block count. My branch spent hundreds of blocks at elevated difficulty (peaking ~7,800 near 300,360, then decaying through the low thousands over several days) — very roughly on the order of 300,000+ work units across ~450 blocks. The network's main branch stayed near difficulty ~40, so ~1,000 blocks contribute only ~40,000 work units. By that arithmetic my minority branch carries roughly an order of magnitude more work despite being lower in height. This is not just my calculation — my node confirms it directly by rejecting the network branch as low-work (log below), so the two independent estimates agree.

The result is that I ended up on a minority branch that forked at ~height around 299054 or 300 067 (can not be sure right now) and carries substantially more accumulated work than the network's main branch (my branch reached high difficulty after the spike; the main branch stayed at low difficulty). Heights: my tip 300447, network tip 301111.

<img width="1583" height="337" alt="Image" src="https://github.com/user-attachments/assets/8f3f17cb-a901-4bff-a375-95953b9475d6" />
## Logs

My node correctly identifies the network's longer branch as lower-work:

```
sending headers (112917 bytes) peer=22
received: headers (96069 bytes) peer=22
Ignoring low-work chain (height=301111) from peer=22
disconnecting peer=22

```

My node sends its (heavier) headers to the peer before disconnecting, so the peer does receive them. However, the network peers do not reorg onto my heavier branch, and my node drops them as low-work. The result is a stable split: each side keeps its own tip and rejects the other.

## Analysis

[...]

PR #171, opened 2026-08-18
test: cover LWMA vs the 144-block anti-DoS floor

bussyjd:
## What

Add a unit test of `GetAntiDoSWorkThreshold`'s tip-relative formula and a short note. No consensus or `net_processing` change.

## Why

Issue #168 reports a public-testnet split: a shorter, heavier LWMA branch and the longer low-diff branch each treat the other as a low-work DoS. Bitcoin's 144-block buffer assumes competing tips have similar per-block work. BTQ's per-block LWMA at 60s spacing can make that false.

Regtest cannot reproduce live swings (`fPowNoRetargeting`, `nLWMAHeight = 300000`). This change only characterizes the arithmetic so the issue can be discussed with numbers.

## Approach

- Replicate the private helper's formula in `src/test/pow_lwma_tests.cpp` (do not friend `PeerManagerImpl`).
- Cases: near-tip fork absorbed by the 144-block buffer; H=450 and r≈0.005 (the observed shape) falls below the threshold; H=450 and r≈1 stays above; `nMinimumChainWork == 0` does not save testnet/regtest.
- Note at `contrib/btq/lwma-antidos-note.md`. Measure before changing 144.

Refs #168

## Testing

```
src/test/test_btq --run_test=pow_lwma_tests
```

Ran on Ubuntu 24.04 (Ryzen 9 5950X): 15 cases, no errors.

## Risks

Low. Test and note only. Does not change the 144-block constant or consensus.

### DIFF
--- a/contrib/btq/lwma-antidos-note.md
+++ b/contrib/btq/lwma-antidos-note.md
@@ -1,0 +1,74 @@
+# LWMA and the 144-block anti-DoS work threshold
+
+Characterization note for [GitHub issue #168](https://github.com/btq-ag/btq-core/issues/168).
+No consensus or net_processing change is proposed here.
+
+## What `GetAntiDoSWorkThreshold` does
+
+`PeerManagerImpl::GetAntiDoSWorkThreshold` in `src/net_processing.cpp`
+computes:
+
+```
+near = tip.nChainWork - min(144 * GetBlockProof(tip), tip.nChainWork)
+threshold = max(near, MinimumChainWork())
+```
+
+Incoming headers (and compact-block / block paths that consult the same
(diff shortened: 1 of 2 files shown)


bussyjd:
Closing and superseding this draft. Independent review found that the mutual low-work rejection claim is mathematically incorrect because each threshold is bounded by its own chainwork. Public block-300000 replay instead shows that N=45 matches the canonical target while N=144 differs, so consensus-version divergence must be investigated first. I will replace this with two focused, independently testable efforts: (1) N=45 versus N=144 activation evidence and (2) production-path headers anti-DoS coverage. Refs #168.

Issue #147, opened 2026-08-05
test: script_assets_test has never run — the corpus URL 404s and the job is not wired into CI


BarneyChambers:
`script_tests/script_assets_test` is the differential script-validation corpus: ~2,244 transactions covering Taproot, tapscript, OP_SUCCESS, witness programs and legacy script, each verified under every consensus flag combination. For a fork that added opcodes and a sigversion to the interpreter it is the single highest-value test in the tree.

It has never executed here, in any environment.

## Why it never runs

**Locally** it self-skips, and the skip is a `BOOST_WARN` rather than a failure, so it is invisible in a green run:

```
test/script_tests.cpp(1798): warning: in "script_tests/script_assets_test":
Variable DIR_UNIT_TEST_DATA unset, skipping script_assets_test
```

That is the entire content of the "1 failed warning" line at the bottom of every unit run.

**In CI** the fetch is broken and the job is not wired up. `ci/test/06_script_b.sh:70` pulls the corpus from:

```
https://github.com/btq-core/qa-assets/raw/main/unit_test_data/script_assets_test.json
```

| URL | status |
| --- | --- |
| `github.com/btq-core/qa-assets` | 404 |
| `github.com/btq-ag/qa-assets` | 404 |
| `github.com/bitcoin-core/qa-assets` | 200 |

The org is `btq-ag`, not `btq-core`, and neither org has a `qa-assets` repo. Separately, `06_script_b.sh` is not referenced by `build-and-address-tests.yml` or `ci.yml`, so even with a working URL nothing would invoke it.

## What it says when you do run it

Ran it against upstream's corpus (8.9 MB, 2,244 cases) on master `d5d1b1e82`:

```
279,079 of 279,341 assertions passed
      262 assertions failed
```

Localised the failures by splitting the corpus by witness version of the spent output:

| bucket | cases | result |
| --- | --- | --- |
| not a witness program | 482 | pass |
| v0 | 100 | pass |
| **v1 (Taproot)** | **1490** | **262 failed** |
| v2 | 14 | pass |
| v3..v16 | 158 | pass |

Then bisected the Taproot bucket down to the individual cases. **11 cases of 1,490 fail, and all 11 have the same cause.**

```
opsuccess/exexecif        opsuccess/bare          tapscript/inputmaxlimit
opsuccess/bigpush         opsuccess/1001inputs    tapscript/pushmaxlimit
opsuccess/unexecif
```

[...]

PR #172, opened 2026-08-18
test: run script_assets_test on a pinned corpus

bussyjd:
## What

Make `script_tests/script_assets_test` actually run. Fetch Bitcoin Core's script-oracle JSON from a **pinned** `bitcoin-core/qa-assets` commit and skip the eleven cases that diverge on purpose.

## Why

The test is the ~2,244-case Taproot/legacy/script differential corpus. It has never run here: `DIR_UNIT_TEST_DATA` is unset locally (`BOOST_WARN` skip), and CI cloned `btq-core/qa-assets`, which 404s.

That file is still the right oracle for the ECDSA/Schnorr/Taproot machine BTQ still ships. It is not a Dilithium or P2MR test. On a live `test_btq` against this JSON: 262 raw failures, all in 11 named cases; after the skips, 0 unexplained failures.

## Approach

- Point lean CI and `ci/test/06_script_b.sh` at
  `bitcoin-core/qa-assets` commit `b33d85102d169b54d966ea315ad81a636680aefa`
  (last change to `script_assets_test.json`, 2025-07-23). Do not track `main`.
- Skip via a GetOp walk of executed scripts for opcodes `0xbb`–`0xbf` (Dilithium reclaimed BIP342 `OP_SUCCESS`), plus comment prefixes `tapscript/inputmaxlimit` / `tapscript/pushmaxlimit` (Bitcoin 520-byte element cap; BTQ allows 15,000).
- Leave the unset-env skip as `BOOST_WARN` so `make check` stays green without the JSON.
- `btq-ag/qa-assets` is TBD. This PR does not create that repo.

Refs #147

## Testing

```
DIR_UNIT_TEST_DATA=<dir-with-json> src/test/test_btq --run_test=script_tests/script_assets_test
```

Ran on Ubuntu 24.04: unpatched 262 failures; patched 11 skips, no errors.

## Risks

Low. No consensus change. CI now depends on a pinned third-party blob remaining at that SHA. If GitHub deletes the raw file, the new lean-CI step fails closed.

### DIFF
--- a/.github/workflows/build-and-address-tests.yml
+++ b/.github/workflows/build-and-address-tests.yml
@@ -69,6 +69,21 @@ jobs:
       - name: Unit tests (make check)
         run: make check
 
+      # Bitcoin Core's script oracle, pinned. Do not use /main — it moves.
+      # Last commit that touched this JSON: 2025-07-23.
+      # Intended long-term home: btq-ag/qa-assets (TBD).
+      - name: script_assets_test (pinned bitcoin-core/qa-assets)
+        env:
+          SCRIPT_ASSETS_COMMIT: b33d85102d169b54d966ea315ad81a636680aefa
+        run: |
+          set -euo pipefail
+          ASSETS_DIR="$RUNNER_TEMP/qa-assets-unit"
+          mkdir -p "$ASSETS_DIR"
+          curl --location --fail \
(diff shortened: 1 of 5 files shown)


augchan42:
Closing as superseded by #173.

#173 contains this PR's commit `00fb1ffb` ("test: run script_assets_test on a pinned corpus") verbatim, plus `4b56c8f8` adding the weekly `qa-assets` pin bump. The two PRs also touch the same five files, so keeping both open only creates a conflict between them.

Reviewing and merging #173 delivers everything here. No work is lost.

PR #173, opened 2026-08-18
ci: add weekly qa-assets pin bump

### LINKED ISSUE #147: test: script_assets_test has never run — the corpus URL 404s and the job is not wired into CI

bussyjd:
## Why

`script_tests/script_assets_test` checks BTQ's script interpreter against Bitcoin Core's shared test data: 2,244 transactions covering legacy script, SegWit v0, Taproot and tapscript, each checked under every combination of consensus flags. BTQ changed the interpreter for Dilithium, so this is the test most likely to catch an accidental difference from Bitcoin's script rules.

It has never run in this repo (#147):

- Locally it skips itself when `DIR_UNIT_TEST_DATA` is unset. The skip is a warning, so `make check` still shows green.
- `ci/test/06_script_b.sh` downloads the data from `github.com/btq-core/qa-assets`, which does not exist. No workflow in `.github/workflows` runs that script anyway.

Run against Bitcoin Core's data, 11 of the 2,244 cases fail. All 11 test rules BTQ changed on purpose:

- BIP342 makes opcodes `0xbb`–`0xbf` `OP_SUCCESS`. BTQ uses them for Dilithium (`IsOpSuccess` in `src/script/script.cpp`).
- Bitcoin limits a stack element to 520 bytes. BTQ allows 15,000, so the cases that expect a 521-byte element to fail now pass.

## What this changes

- `contrib/qa-assets.pin` holds one commit of `bitcoin-core/qa-assets` (`b33d851`, the last commit that changed `script_assets_test.json`). CI downloads the file from that commit, never from `main`, so the test data changes only through a reviewed PR.
- `build-and-address-tests.yml` gets a step that downloads the pinned file and runs `script_assets_test`.
- `src/test/script_tests.cpp` skips the 11 cases above. A case is skipped if one of its executed scripts contains opcode `0xbb`–`0xbf` as an instruction (found with `GetOp`, not a byte search), or if its name starts with `tapscript/inputmaxlimit` or `tapscript/pushmaxlimit`. Every other case must pass.
- `.github/workflows/qa-assets-bump.yml` runs every Monday. If the last upstream commit that changed `script_assets_test.json` differs from the pin, it opens a PR that moves the pin. It never merges, and it skips when a branch for that commit already exists.
- `ci/test/06_script_b.sh`, `doc/fuzzing.md` and `doc/audit-briefing.md` point at `bitcoin-core/qa-assets` at the pinned commit.

qa-assets `main` gets new fuzz inputs every week, but `script_assets_test.json` last changed on 2025-07-23. Watching only that file keeps bump PRs rare.

## Not changed

No node, wallet or consensus code. The diff is CI config, one CI script, the pin file, docs and `src/test/script_tests.cpp`.

## After merge

Add a repo secret `QA_ASSETS_BUMP_TOKEN`: a personal access token with contents and workflow scope. PRs opened with the default `GITHUB_TOKEN` do not start other workflows, so without the secret a bump PR has no CI.

## If the test goes red later

A bump PR fails when Bitcoin adds cases that touch a rule BTQ changed, or when BTQ's interpreter has drifted. Add a skip with the reason, or fix the interpreter. Do not merge a red bump.

## Testing

- CI on this PR: build, `make check` and `script_assets_test` pass. `test each commit` failed because that run used the Sep 10 `ci.yml`, which ran the full functional suite. Those failures are the ones already on master (#104).
- Locally on master + this PR: `script_assets_test` passes with 11 cases skipped. The pin matches the latest upstream commit for the file.

Closes #147. On the issue's fourth item: the skip when the data is missing stays a warning, so `make check` passes without the file. The CI step downloads it with `curl --fail` under `set -e`, so a failed download fails CI instead of skipping the test.

Supersedes #172, which held the first commit of this PR.

### DIFF
--- a/.github/workflows/build-and-address-tests.yml
+++ b/.github/workflows/build-and-address-tests.yml
@@ -69,14 +69,11 @@ jobs:
       - name: Unit tests (make check)
         run: make check
 
-      # Bitcoin Core's script oracle, pinned. Do not use /main — it moves.
-      # Last commit that touched this JSON: 2025-07-23.
-      # Intended long-term home: btq-ag/qa-assets (TBD).
+      # Bitcoin Core's script oracle. SHA lives in contrib/qa-assets.pin.
       - name: script_assets_test (pinned bitcoin-core/qa-assets)
-        env:
-          SCRIPT_ASSETS_COMMIT: b33d85102d169b54d966ea315ad81a636680aefa
         run: |
           set -euo pipefail
+          SCRIPT_ASSETS_COMMIT=$(grep -E '^[0-9a-f]{40}$' contrib/qa-assets.pin)
           ASSETS_DIR="$RUNNER_TEMP/qa-assets-unit"
           mkdir -p "$ASSETS_DIR"
           curl --location --fail \
--- a/.github/workflows/build-and-address-tests.yml
+++ b/.github/workflows/build-and-address-tests.yml
@@ -69,6 +69,21 @@ jobs:
       - name: Unit tests (make check)
         run: make check
 
+      # Bitcoin Core's script oracle, pinned. Do not use /main — it moves.
+      # Last commit that touched this JSON: 2025-07-23.
+      # Intended long-term home: btq-ag/qa-assets (TBD).
+      - name: script_assets_test (pinned bitcoin-core/qa-assets)
+        env:
+          SCRIPT_ASSETS_COMMIT: b33d85102d169b54d966ea315ad81a636680aefa
+        run: |
+          set -euo pipefail
+          ASSETS_DIR="$RUNNER_TEMP/qa-assets-unit"
+          mkdir -p "$ASSETS_DIR"
+          curl --location --fail \
+            "https://raw.githubusercontent.com/bitcoin-core/qa-assets/${SCRIPT_ASSETS_COMMIT}/unit_test_data/script_assets_test.json" \
+            -o "$ASSETS_DIR/script_assets_test.json"
+          DIR_UNIT_TEST_DATA="$ASSETS_DIR" src/test/test_btq --run_test=script_tests/script_assets_test
+
       - name: Smoke test - wallet + address generation (regtest)
         # Self-contained smoke test, on purpose NOT using the upstream
         # functional-test framework:
--- a/.github/workflows/qa-assets-bump.yml
+++ b/.github/workflows/qa-assets-bump.yml
@@ -1,0 +1,88 @@
+# Copyright (c) 2026 The BTQ Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#
+# Watch bitcoin-core/qa-assets for commits that touch script_assets_test.json.
+# If that path moved, open a pin PR. Do not follow repo main (fuzz corpora).
+#
+# GITHUB_TOKEN pushes do not start other workflows. The pin PR will not run
+# lean CI unless QA_ASSETS_BUMP_TOKEN (a PAT with workflow + contents) is set.
+
+name: qa-assets pin bump
+
+on:
+  schedule:
+    - cron: "27 7 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  bump:
+    if: github.repository == 'btq-ag/btq-core'
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Resolve latest JSON-touching commit
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          NEW=$(gh api \
+            "repos/bitcoin-core/qa-assets/commits?path=unit_test_data/script_assets_test.json&per_page=1" \
+            --jq '.[0].sha')
+          CUR=$(grep -E '^[0-9a-f]{40}$' contrib/qa-assets.pin)
+          echo "new=$NEW" >> "$GITHUB_ENV"
+          echo "cur=$CUR" >> "$GITHUB_ENV"
+          echo "remote $NEW"
+          echo "pin    $CUR"
+
+      - name: Open pin PR
+        if: env.new != env.cur
+        env:
+          GH_TOKEN: ${{ secrets.QA_ASSETS_BUMP_TOKEN || secrets.GITHUB_TOKEN }}
+          NEW: ${{ env.new }}
+        run: |
+          set -euo pipefail
+          SHORT="${NEW:0:12}"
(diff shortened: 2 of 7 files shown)


augchan42:
## Independent review — revised

The path-scoped pinning design is right, and selecting the last commit that touched the oracle JSON is correct. One automation bug stands.

The workflow skips only an **open** existing PR:

```
gh pr list --head "$BRANCH" --state open
```

The branch name is deterministic per SHA (`ci/qa-assets-pin-<sha>`). If a PR for that SHA is opened and then closed, the remote branch survives. On the next run the check passes, `git checkout -B` creates a divergent local branch from a fresh shallow checkout, and the plain `git push` is rejected as non-fast-forward. That SHA then fails on every run.

Either query the remote branch instead of the PR state, or add a run-unique suffix to the branch name.


---

augchan42:
## Follow-up review — the branch check is fixed

`155a4f65b` checks for the remote branch with `git ls-remote --exit-code --heads`. A branch left over from a closed PR no longer makes every later run fail.

Tested locally on master + this PR: `script_assets_test` passes against the pinned `script_assets_test.json` (`b33d851`, still the latest qa-assets commit that touches that file). It skips 11 cases: `opsuccess/bare`, `opsuccess/bigpush` and `opsuccess/1001inputs` (two each), `opsuccess/unexecif`, `opsuccess/exexecif`, `tapscript/inputmaxlimit` (two) and `tapscript/pushmaxlimit`. That matches the set the PR describes. The PR merges cleanly with master.

Notes, none blocking:

1. CI has not run on this PR. Both runs from the fork are waiting for maintainer approval, so the new `script_assets_test` step in `build-and-address-tests.yml` has never run in CI. A maintainer needs to approve them before merge.
2. The repo has no `QA_ASSETS_BUMP_TOKEN` secret. Pin PRs opened with `GITHUB_TOKEN` will not start lean CI, so a bump PR shows no checks unless someone closes and reopens it. The workflow comment says this; the secret needs to be set.
3. (low) If the push succeeds but `gh pr create` fails, the branch exists with no PR, and every later run for that SHA exits 0 with "Branch already exists". Actions is allowed to create PRs in this repo today, so this only matters if that setting changes. When the branch exists, checking `gh pr list --head "$BRANCH" --state all` and opening the PR if there is none would cover it.
4. (low) The fuzz path in `ci/test/06_script_b.sh` now clones all of `bitcoin-core/qa-assets`, fuzz corpora and history included, instead of `--depth=1`. No current CI job runs that path. `git init` followed by `git fetch --depth=1 origin <sha>` would keep it shallow.

---

augchan42 (approved):
Approving. CI passed after the fork runs were approved: the build, `make check` and the new `script_assets_test` step against the pinned corpus all passed.

`test each commit` failed, but not because of this PR. That run used the Sep 10 `ci.yml`, which still ran the full functional suite. The failures are functional tests that already fail on master (#104), such as `rpc_psbt.py`, `wallet_miniscript.py` and `p2p_segwit.py`. This PR changes no node or wallet code. Current master no longer runs the functional suite in that job.

Setting the `QA_ASSETS_BUMP_TOKEN` secret is still needed so bump PRs get CI.

Issue #174, opened 2026-08-18
consensus: reconstruct testnet LWMA activation version split


bussyjd:
## Summary

The public testnet split reported in #168 began at the configured LWMA activation height. Before changing consensus or attributing the incident to header-work filtering, reconstruct which difficulty rule each deployed binary and branch enforced.

## Verified public evidence

- The official v0.4.2 testnet source and Linux binary use an LWMA window of 45.
- The later v0.4.4 testnet release uses a window of 144 and identifies that window change as consensus-affecting.
- The explorer header at height 300000 commits to `nBits=1c04675d`.
- An independent replay of explorer headers 299855 through 299999 yields `1c04675d` for N=45 and `1c055bf4` for N=144.

This establishes the rule followed by the explorer branch. It does not establish the reporting miner node version or prove the complete incident root cause.

## Evidence still required

- Exact reporting-node version, subversion, commit or build provenance, and binary SHA256.
- Raw headers and verbose header JSON at heights 299999 through 300001 from both branches.
- The first differing header and rejection reason on each side.
- Remote-node logs, not only the local low-work log.
- Independent confirmation of the canonical header corpus from a synced node in addition to the explorer.

## Proposed test-only effort

- Pin the public activation header corpus with explicit provenance.
- Exercise production `GetNextWorkRequired` for N=45 and N=144.
- Add an offline arbitrary-precision oracle that independently checks serialization, linkage, PoW, compact targets, and both exact results.
- Make no consensus or network-parameter change.

## Completion gate

- Results reproduce from the pinned fixture without network access.
- Limitations are explicit: the test does not select a canonical recovery policy and does not close #168.
- Two exact-head public ACKs, including a difficulty-domain reviewer, plus green relevant CI.

Refs #168 and #143.

bussyjd:
@unkn0wn747, for incident evidence (not a code-review ACK), could you attach:

- the reporting node `btqd --version`, `getnetworkinfo.subversion`, commit/build provenance, and binary SHA256;
- raw and verbose headers at heights 299999 through 300001 from the reporting branch;
- the first rejection reason and available remote-node logs.

PR #176 intentionally remains Draft and does not claim root-cause closure without those artifacts.

unkn0wn747:
>> the reporting node btqd --version, getnetworkinfo.subversion, commit/build provenance, and binary SHA256;

My node code was master cloned on August 8 (corresponding to v0.4.2 plus many fixes merged to master by that date).
[...]

PR #176, opened 2026-08-18
test: pin N=45 and N=144 activation targets

### LINKED ISSUE #168: Observation: heavy-work minority fork does not propagate; possible relevance to mainnet genesis

bussyjd:
## Summary

Draft forensic-evidence PR for the public testnet LWMA activation at height 300000.

The pinned explorer header at height 300000 has `nBits=1c04675d`. Replaying the preceding canonical headers through production `GetNextWorkRequired` yields:

- N=45: weighted solvetime 4423, `1c04675d` (matches the observed header)
- N=144: weighted solvetime 54294, `1c055bf4` (does not match)

Current BTQ testnet parameters remain N=144. This PR therefore exposes an unresolved historical-compatibility problem; it deliberately does not choose or implement a consensus remedy and does not close the incident.

Refs #174, #168, and #143.

## Motivation

PR #171 attributed the reported split to mutual low-work filtering. That explanation is not valid. Before any consensus or recovery decision, the project needs a reproducible record of which LWMA target the observed activation history commits to.

The public evidence still does not identify the reporting miner binary or its competing height-300000 header. Those remain required incident artifacts.

## Approach

- Pin 146 raw 80-byte public testnet headers, heights 299855 through 300000, with exact per-height endpoint provenance.
- Verify serialization exhaustion, SHA256d hashes, linkage, PoW, compact targets, and chainwork increments.
- Call production `GetNextWorkRequired` with copied testnet parameters for N=45 and N=144.
- Run a structurally independent, offline Python oracle using arbitrary-precision integers.
- Run that oracle explicitly in the active lean CI workflow after `make check`.

Synthetic-only vectors and the in-tree copied LWMA reference were rejected as incident evidence because neither proves what the observed public branch committed to. A regtest functional simulation is also out of scope: regtest currently returns early through `fPowNoRetargeting`.

## Provenance and limitations

- Source: `https://explorer.bitcoinquantum.com/api/v1/block/{height}`
- Retrieval date: 2026-08-18
- All 146 responses were re-fetched, required `canonical=true`, and reconstructed byte-for-byte.
- Fixture SHA256: `ebf1fe461d17c3fd0d1c8f4667f41984eaf1358e8d8cf4333e39b9e415bce222`

The explorer is still a single mutable external source. The corpus is self-consistent and independently checked, but an independent synced-node attestation is still required. The first chainwork value is not replayed from genesis, and merkle roots are not recomputed without full block bodies.

## Testing

Ubuntu 22.04 aarch64, GCC, C++17:

```bash
python3 test/lint/lint-btq-lwma-activation.py
make -C src -j2 test/test_btq-pow_lwma_tests.o
src/test/test_btq --run_test=pow_lwma_tests/lwma_testnet_activation_window_divergence --log_level=test_suite
src/test/test_btq --run_test=pow_lwma_tests --log_level=message
python3 -m py_compile test/lint/lint-btq-lwma-activation.py
git diff --check e2d19e06b..HEAD
```

All focused commands passed. The full `pow_lwma_tests` suite ran 15 cases with no errors. Workflow YAML parsing also passed.

Broad `all-lint.py` is not claimed green; the repository has unrelated baseline lint/type failures.

## Bitcoin baseline

Pinned baseline: Bitcoin Core 26.0 commit `44d8b13c81e5276eb610c99f227a4d090cc532f6`.

Bitcoin 26 has no LWMA, so an N=45/N=144 numerical differential is not applicable. This PR must not be represented as a Bitcoin-baseline differential test.

## Risk and compatibility

**Risk: low for runtime behavior, high for interpretation.** This changes test data, tests, and CI only. It changes no consensus, network, wallet, or RPC behavior. The evidence must not be used to imply that current source is compatible with the observed history; it is not.

No release note is needed for this test-only change.

## Audit-matrix scope

This is a scoped evidence increment for C03. It does not satisfy C03: wrong-bits node rejection, activation reorg/reindex coverage, a reviewed compatibility rule, and additional independent evidence remain missing. It does not close #168.

## Review gate

Keep Draft until:

- a difficulty reviewer independently re-derives both compact targets;
- a second reviewer re-fetches or independently attests the header corpus;
- two exact-head public ACKs are recorded;
- the relevant CI commands run and pass on this exact head.

Any force-push invalidates prior ACKs until reviewers inspect the range-diff.


bussyjd:
Review requested for exact head `cd9f7ab3538c8bad667e1a4528d5dd37f68e530c`.

- @augchan42: please independently re-derive the N=45 and N=144 compact targets and review the historical parameter interpretation.
- @BarneyChambers: please review the consensus-compatibility implications and the explicit non-closure of #168/C03.
- @Oskii: please review fixture/test ownership, production-path relevance, and the active-CI oracle step.

Please record any final approval as:

```text
Tested ACK cd9f7ab3538c8bad667e1a4528d5dd37f68e530c

Reviewed:
- ...

Executed:
- command -> result
```

A maintainer will need to add the formal reviewer assignments because this fork account lacks upstream `RequestReviewsByLogin` permission. Any new head requires renewed ACKs.

bussyjd:
CI gate status for exact head `cd9f7ab3538c8bad667e1a4528d5dd37f68e530c`: both workflows ended `action_required` before running any jobs.

- build-and-address-tests: https://github.com/btq-ag/btq-core/actions/runs/32193858143
- CI: https://github.com/btq-ag/btq-core/actions/runs/32193858144

An upstream maintainer must approve the fork workflows. This is not a pass or a failure; the Draft is not CI-green.

Issue #175, opened 2026-08-18
p2p: verify header-work filtering boundaries


bussyjd:
## Summary

The inherited low-work header filter needs BTQ-specific production-path coverage. The current functional test also hardcodes the Bitcoin regtest genesis hash, so part of its state assertion is invalid for BTQ.

This effort characterizes existing P2P behavior. It does not change consensus, tune policy, or claim to explain #168.

## Proposed scope

- Extract the existing threshold arithmetic into a production-called helper without changing behavior.
- Cover saturation, minimum-chainwork dominance, and variable current-tip proof in the owning headers-sync unit suite.
- Repair the functional test to derive the BTQ regtest genesis dynamically.
- Through the actual P2P path, verify that a chain exactly at the work threshold is accepted, one below is ignored, and the sending peer remains connected.
- Verify directionality: the higher-work side can reject a lower-work branch while the lower-work side accepts the higher-work branch.
- Compare the shared inherited behavior with the pinned Bitcoin Core 26.0 baseline where parameters are equivalent.

## Explicit non-goals

- No LWMA incident-root-cause claim.
- No mainnet attack corpus or throughput reproducer.
- No minimum-chainwork, inherited 144-block buffer, or headers-sync calibration change.
- No claim that this closes the broader audit matrix rows.

Any suspected mainnet abuse scenario or launch-parameter weakness must be handled through the private security process, not this public issue.

## Completion gate

- Tests call production code and assert block-index or peer state, not only copied arithmetic or log text.
- Unit and focused functional tests pass on the exact reviewed head.
- Two exact-head public ACKs, including network and test-domain review, plus green relevant CI.

Refs #168 for incident context only.

PR #177, opened 2026-08-18
test: cover header-work filtering through P2P

### LINKED ISSUE #168: Observation: heavy-work minority fork does not propagate; possible relevance to mainnet genesis

bussyjd:
## Summary

Draft coverage/refactor PR for the inherited header-work anti-DoS path.

It extracts the existing threshold expression into a production-called helper, verifies its BTQ-specific work vectors, repairs the stale Bitcoin regtest-genesis assertion, and exercises exact acceptance/rejection boundaries through `PeerManager` over P2P.

This is intentionally a behavior-preserving characterization. It does not fix or explain the testnet incident.

Refs #175. #168 is context only.

## Motivation

The previous PR copied a private threshold helper into a unit test and used it to support a mutual-rejection incident claim. Such a test can pass even if production changes, and mutual threshold rejection is impossible when each threshold is bounded by its own chainwork.

The owning inherited functional test also hardcoded Bitcoin regtest genesis, so part of its state assertion was invalid for BTQ.

## Approach

- Extract `CalculateAntiDoSWorkThreshold` into `headerssync` and keep `PeerManagerImpl::GetAntiDoSWorkThreshold` as its production caller under the same `cs_main` scope.
- Preserve the exact inherited expression: `max(W - min(144*p, W), M)`.
- Add unit vectors for subtraction saturation, minimum-chainwork dominance, and variable current-tip work.
- Derive the regtest genesis hash dynamically and move the floor to exact equality at `0x20`.
- Through actual P2P header processing, assert below-threshold absence from the block index, equality acceptance, asymmetric high/low behavior, and peer connection retention.
- Run the targeted functional test in the active lean CI workflow.

Regtest has constant work. Variable tip-work arithmetic is therefore owned by the production-helper unit vectors; the functional test owns network-path behavior.

## Exact behavior checked

- Regtest chainwork `0x1e` is ignored below a `0x20` floor.
- Equality at `0x20` is accepted because production compares `<`, not `<=`.
- A higher-work node filters the lower-work fork; the lower-work node accepts the higher-work fork.
- A header exactly at the dynamic tip-relative threshold is stored; one regtest block-work below is not.
- Low-work filtering does not disconnect the announcing peers.

These outcomes refute mutual low-work filtering as a self-contained stable-split explanation. They do not identify the root cause of #168.

## Testing

Ubuntu 22.04 aarch64, GCC, C++17:

```bash
src/test/test_btq --run_test=headers_sync_chainwork_tests
python3 test/functional/test_runner.py p2p_headers_sync_with_minchainwork.py --jobs=1
python3 -m py_compile test/functional/p2p_headers_sync_with_minchainwork.py
git diff --check e2d19e06b..HEAD
```

Results:

- clean `btqd` and `test_btq` build passed;
- both unit cases passed;
- exact workflow functional command passed;
- four deterministic seed/portseed repeats (101, 202, 303, 404) passed;
- targeted flake8 and mypy passed;
- workflow YAML parsing passed.

Broad repository lint is not claimed green; the wrapper currently reports unrelated baseline type failures.

## Bitcoin 26 differential

Pinned baseline: `44d8b13c81e5276eb610c99f227a4d090cc532f6`.

The same modified functional scenario was run against a clean Bitcoin Core 26.0 worktree with no baseline production file changed, seed/portseed 606. It completed successfully with the same below/equality/directionality/connection outcomes. The threshold production code at the BTQ base is inherited from that baseline.

## Risk and compatibility

**Risk: low, network code touched.** The production edit is an algebraically identical helper extraction with unchanged lock ownership and configured floor. No consensus, wire-format, policy value, wallet, RPC, or configuration behavior changes.

The public scope deliberately excludes mainnet attack fixtures, throughput or memory-amplification reproducers, minimum-chainwork changes, the inherited 144-block buffer choice, and headers-sync parameter recalibration. Any suspected mainnet weakness belongs in the private security process.

No release note is needed for this coverage/refactor.

## Audit-matrix scope

This is a scoped evidence increment for C19. It does not satisfy C19: a replacement BTQ header corpus, invalid LWMA/P2MR relay vectors, calibration review, and independent external review remain missing. It does not satisfy C03 or close #168.

## Review gate

Keep Draft until:

- a network/DoS reviewer traces the production call path and records a Tested ACK;
- a PoW reviewer independently checks the work arithmetic;
- a test-framework reviewer confirms state assertions and cleanup;
- two exact-head public ACKs and relevant green CI are recorded.

Any force-push invalidates prior ACKs until reviewers inspect the range-diff.

### DIFF
--- a/.github/workflows/build-and-address-tests.yml
+++ b/.github/workflows/build-and-address-tests.yml
@@ -69,6 +69,9 @@ jobs:
       - name: Unit tests (make check)
         run: make check
 
+      - name: Header-work P2P regression
+        run: python3 test/functional/test_runner.py p2p_headers_sync_with_minchainwork.py --jobs=1
+
       - name: Smoke test - wallet + address generation (regtest)
         # Self-contained smoke test, on purpose NOT using the upstream
         # functional-test framework:
--- a/src/headerssync.cpp
+++ b/src/headerssync.cpp
@@ -9,6 +9,11 @@
 #include <util/check.h>
 #include <util/vector.h>
 
+#include <algorithm>
+
+//! Number of tip-work-equivalent blocks kept as a near-tip relay buffer.
+constexpr uint32_t ANTI_DOS_HEADER_WORK_BUFFER{144};
+
 // The two constants below are computed using the simulation script in
 // contrib/devtools/headerssync-params.py.
 
@@ -22,6 +27,17 @@ constexpr size_t REDOWNLOAD_BUFFER_SIZE{14441}; // 14441/606 = ~23.8 commitments
 // Our memory analysis assumes 48 bytes for a CompressedHeader (so we should
 // re-calculate parameters if we compress further)
 static_assert(sizeof(CompressedHeader) == 48);
+
+arith_uint256 CalculateAntiDoSWorkThreshold(
+    const arith_uint256& tip_chain_work,
+    const arith_uint256& tip_block_work,
+    const arith_uint256& minimum_chain_work)
+{
+    const arith_uint256 buffered_work = std::min<arith_uint256>(
+        tip_block_work * ANTI_DOS_HEADER_WORK_BUFFER,
+        tip_chain_work);
+    return std::max<arith_uint256>(tip_chain_work - buffered_work, minimum_chain_work);
+}
 
 HeadersSyncState::HeadersSyncState(NodeId id, const Consensus::Params& consensus_params,
         const CBlockIndex* chain_start, const arith_uint256& minimum_required_work) :
--- a/src/headerssync.h
+++ b/src/headerssync.h
@@ -52,6 +52,18 @@ struct CompressedHeader {
     };
 };
 
+/**
+ * Calculate the minimum chain work required before headers may be stored.
+ *
+ * The 144-block buffer count is header-relay policy inherited from Bitcoin
+ * Core; the resulting work threshold still scales with the current tip's
+ * block work.
+ */
+arith_uint256 CalculateAntiDoSWorkThreshold(
+    const arith_uint256& tip_chain_work,
+    const arith_uint256& tip_block_work,
+    const arith_uint256& minimum_chain_work);
+
 /** HeadersSyncState:
  *
  * We wish to download a peer's headers chain in a DoS-resistant way.
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -2464,15 +2464,18 @@ bool PeerManagerImpl::CheckHeadersPoW(const std::vector<CBlockHeader>& headers, 
 
 arith_uint256 PeerManagerImpl::GetAntiDoSWorkThreshold()
 {
-    arith_uint256 near_chaintip_work = 0;
+    arith_uint256 tip_chain_work{0};
+    arith_uint256 tip_block_work{0};
     LOCK(cs_main);
     if (m_chainman.ActiveChain().Tip() != nullptr) {
         const CBlockIndex *tip = m_chainman.ActiveChain().Tip();
-        // Use a 144 block buffer, so that we'll accept headers that fork from
-        // near our tip.
-        near_chaintip_work = tip->nChainWork - std::min<arith_uint256>(144*GetBlockProof(*tip), tip->nChainWork);
+        tip_chain_work = tip->nChainWork;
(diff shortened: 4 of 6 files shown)


bussyjd:
Review requested for exact head `94b99e6709276c1012c3c707ab40e10196d82c7d`.

- @BarneyChambers: please trace the production P2P call path, lock scope, and behavior-neutral extraction.
- @augchan42: please independently check the floor and tip-relative work arithmetic and the directionality claim.
- @Oskii: please run/review the functional state assertions, peer cleanup, seeded repeats, and targeted CI gate.

Please record any final approval as:

```text
Tested ACK 94b99e6709276c1012c3c707ab40e10196d82c7d

Reviewed:
- ...

Executed:
- command -> result
```

A maintainer will need to add the formal reviewer assignments because this fork account lacks upstream `RequestReviewsByLogin` permission. Any new head requires renewed ACKs.

bussyjd:
CI gate status for exact head `94b99e6709276c1012c3c707ab40e10196d82c7d`: both workflows ended `action_required` before running any jobs.

- build-and-address-tests: https://github.com/btq-ag/btq-core/actions/runs/32193881276
- CI: https://github.com/btq-ag/btq-core/actions/runs/32193881409

An upstream maintainer must approve the fork workflows. This is not a pass or a failure; the Draft is not CI-green.

PR #181, opened 2026-08-25
wallet: refuse trivial P2MR leaves unless allowed

BarneyChambers:
## Summary
- `CreateP2MR` / `FundP2MR` reject empty and OP_TRUE leaves by default. `script: "51"` is consensus-valid and anyone-can-spend; RPC help used to teach that tree.
- Tests and explicit anyone-can-spend callers pass `allow_trivial_leaves`. `getnewp2mraddress` (param 2) and `sendtop2mr` (param 6) expose the same flag.
- `sendtop2mr` maps that reject to `RPC_WALLET_ERROR` (-4), not insufficient funds.

Closes the C001 / E003 items from the BTQ-Core final report.

## Test plan
- [ ] `src/test/test_btq -t p2mr_tests` (or the wallet unit binary that holds `create_rejects_trivial_leaves_by_default`)
- [ ] `test/functional/feature_p2mr_rpc.py --descriptors`
- [ ] `test/functional/wallet_bip360_send_paths.py`
- [ ] Default `sendtop2mr` with `script: "51"` returns -4
- [ ] Same call with `allow_trivial_leaves=true` still funds

### DIFF
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -340,7 +340,8 @@ public:
 
     //! Create and persist a new P2MR destination.
     virtual util::Result<WalletP2MRCreated> createP2MR(const std::vector<WalletP2MRTreeLeaf>& leaves,
-                                                      const std::string& label) = 0;
+                                                      const std::string& label,
+                                                      bool allow_trivial_leaves = false) = 0;
 
     //! Create + persist + fund a P2MR destination atomically.
     virtual util::Result<WalletP2MRFunded> fundP2MR(const std::vector<WalletP2MRTreeLeaf>& leaves,
--- a/src/qt/p2mrcontroller.cpp
+++ b/src/qt/p2mrcontroller.cpp
@@ -80,6 +80,7 @@ CAmount P2MRController::totalBalance(int min_depth)
 
 bool P2MRController::createVault(const std::vector<interfaces::WalletP2MRTreeLeaf>& leaves,
                                  const QString& label,
+                                 bool allow_trivial_leaves,
                                  interfaces::WalletP2MRCreated& out,
                                  QString& error)
 {
@@ -87,7 +88,7 @@ bool P2MRController::createVault(const std::vector<interfaces::WalletP2MRTreeLea
         error = tr("Wallet model unavailable");
         return false;
     }
-    auto res = m_wallet_model->wallet().createP2MR(leaves, label.toStdString());
+    auto res = m_wallet_model->wallet().createP2MR(leaves, label.toStdString(), allow_trivial_leaves);
     if (!res) {
         error = FromBilingual(util::ErrorString(res));
         return false;
--- a/src/qt/p2mrcontroller.h
+++ b/src/qt/p2mrcontroller.h
@@ -75,6 +75,7 @@ public:
     /** Create a new P2MR address. Returns descriptive error on failure. */
     bool createVault(const std::vector<interfaces::WalletP2MRTreeLeaf>& leaves,
                      const QString& label,
+                     bool allow_trivial_leaves,
                      interfaces::WalletP2MRCreated& out,
                      QString& error);
 
--- a/src/qt/p2mrwizards.cpp
+++ b/src/qt/p2mrwizards.cpp
@@ -312,7 +312,8 @@ void P2MRNewVaultDialog::accept()
                 .arg(m_created_address, QString::fromStdString(funded.txid.GetHex())));
     } else {
         interfaces::WalletP2MRCreated created;
-        if (!m_controller->createVault(leaves, label, created, error)) {
+        const bool allow_trivial_leaves = tpl == P2MRController::TreeTemplate::OpTrue;
+        if (!m_controller->createVault(leaves, label, allow_trivial_leaves, created, error)) {
             QMessageBox::critical(this, windowTitle(), error);
             return;
         }
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -306,9 +306,11 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "addnode", 2, "v2transport" },
     // P2MR wallet RPCs
     { "getnewp2mraddress", 0, "tree" },
+    { "getnewp2mraddress", 2, "allow_trivial_leaves" },
     { "sendtop2mr", 0, "tree" },
     { "sendtop2mr", 1, "amount" },
     { "sendtop2mr", 5, "subtractfeefromamount" },
+    { "sendtop2mr", 6, "allow_trivial_leaves" },
     { "createp2mrspend", 2, "amount" },
     { "createp2mrspend", 3, "fee" },
     // Dilithium wallet RPCs
--- a/src/wallet/interfaces.cpp
+++ b/src/wallet/interfaces.cpp
@@ -612,10 +612,12 @@ public:
     }
     util::Result<interfaces::WalletP2MRCreated> createP2MR(
         const std::vector<interfaces::WalletP2MRTreeLeaf>& leaves,
-        const std::string& label) override
+        const std::string& label,
+        bool allow_trivial_leaves) override
     {
         LOCK(m_wallet->cs_wallet);
-        auto res = CreateP2MR(*m_wallet, FromInterfaceLeaves(leaves), label);
+        auto res = CreateP2MR(*m_wallet, FromInterfaceLeaves(leaves), label,
+                              /*add_to_address_book=*/true, allow_trivial_leaves);
         if (!res) return util::Error{util::ErrorString(res)};
         return ToInterfaceCreated(*res);
     }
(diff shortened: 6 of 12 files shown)


augchan42:
## Independent review — the guard misses the cases it's named for

### [high] The predicate catches only empty and *exactly* `OP_TRUE`

`src/wallet/p2mr.cpp:341`:

```cpp
bool IsTrivialP2MRLeaf(const P2MRTreeLeaf& leaf)
{
    return leaf.script.empty() || IsOpTrueLeaf(leaf);
}
```

and `IsOpTrueLeaf` (line 210) requires:

```cpp
leaf.leaf_version == TAPROOT_LEAF_TAPSCRIPT &&
leaf.script.size() == 1 &&
leaf.script[0] == OP_TRUE
```

So the check is an exact single-byte match. It does not catch:

- `OP_2` (`52`) or any other truthy constant — succeeds with no authorization
- `OP_DROP OP_TRUE` (`7551`) — anyone can spend by supplying one stack item
- any `OP_SUCCESSx` — succeeds at consensus; standard-policy discouragement does not protect against a miner or direct block inclusion
- every nonempty **unknown leaf version** — not `0xc0`, so `IsOpTrueLeaf` returns false regardless of content

The gap is already demonstrated in-tree. `test/functional/feature_p2mr_rpc.py:101-106`:

```python
needs_stack_tree = [{
    "depth": 0,
    "leaf_version": LEAF_VERSION_TAPSCRIPT,
    "script": "7551",  # OP_DROP OP_TRUE
}]
needs_stack = wallet.sendtop2mr(needs_stack_tree, Decimal("0.1"), "rpc-p2mr-needs-stack")
```

No `allow_trivial_leaves` argument — compare line 57, which does pass it for the `"51"` tree. So the default path funds an effectively anyone-can-spend output while the new guard reports it is preventing exactly that. A user who reads the error message for `51` will reasonably assume `7551` was also checked.

**Fix:** invert the test. Accept only supported leaf versions and scripts that provably require cryptographic authorization; require the opt-out for unknown versions and unrecognized scripts. At minimum reject truthy constants, stack-consuming constructions like `OP_DROP OP_TRUE`, and `OP_SUCCESSx`. Add cases for `52`, `7551`, an `OP_SUCCESSx`, and a non-`0xc0` leaf version.

### [medium] The Qt "OP_TRUE leaf (regtest only)" template now always fails

`src/wallet/p2mr.h:151`. The `allow_trivial_leaves` boolean was added to `CreateP2MR` and `FundP2MR` with a `false` default, but not plumbed through `interfaces::Wallet::createP2MR` / `fundP2MR`, `WalletImpl`, or `P2MRController`. The Qt dialog deliberately offers that template and calls `wallet().createP2MR(...)` without the new argument, so selecting it now returns the rejection with no way to opt in.

**Fix:** thread the boolean through the interface, implementation and controller, and pass `true` only behind an explicit regtest/test-only confirmation in the dialog. Add coverage for the interface path — the current tests exercise only the direct wallet call.

### Scope note

This is wallet-side only — it changes what the wallet will *construct*, not what a node will accept or relay. Worth stating explicitly in the commit message, since "refuse trivial leaves" reads like a policy change and isn't one. Existing outputs remain spendable.

augchan42:
## Follow-up review — the guard still does not hold

The opt-out plumbing is right, and the RPC and test coverage around it look good. The guard itself does not do what it is named for.

### High

`src/wallet/p2mr.cpp:341`:

```cpp
return leaf.script.empty() || IsOpTrueLeaf(leaf);
```

`IsOpTrueLeaf` requires exactly one byte equal to `OP_TRUE` at `TAPROOT_LEAF_TAPSCRIPT`. Everything else that spends without authorization goes straight through: `OP_2`, `OP_DROP OP_TRUE`, any `OP_SUCCESS` opcode, and any nonempty script under a leaf version we do not recognise.

Your own test shows it. `test/functional/feature_p2mr_rpc.py:111` funds the tree `7551` — `OP_DROP OP_TRUE` — with no `allow_trivial_leaves`, and it succeeds.

I do not think enumerating opcodes gets us there; the list has no natural end. The question worth settling first is what rule we are actually enforcing. Two options:

1. **Allowlist.** A leaf is fundable only if `IsDilithiumLeafSpendable` recognises it, or the caller passes the opt-out. Strict and easy to reason about. It may reject legitimate custom trees.
2. **Require a signature check.** Reject any leaf whose script contains no `OP_CHECKSIGDILITHIUM` or equivalent, and reject unknown leaf versions and `OP_SUCCESS` outright.

I lean towards the allowlist, since the wallet already has to understand a leaf to spend from it. What do you think?

### Medium

The Qt template is unreachable. `src/qt/p2mrwizards.cpp:190` offers "OP_TRUE leaf (regtest only)", but `createP2MR` at `src/interfaces/wallet.h:342` takes only `(leaves, label)` — there is no `allow_trivial_leaves` on the interface, and `src/wallet/interfaces.cpp` calls through with the default `false`. That template always fails. Either carry the flag through the interface or drop the template.


---

augchan42:
**Review finding (High): allowlist breaks `createdilithiummultisig`**

`CreateP2MR()` now rejects any leaf where `IsDilithiumLeafSpendable()` returns false, unless `allow_trivial_leaves` is set (`src/wallet/p2mr.cpp:690`).

`createdilithiummultisig` (`src/wallet/rpc/dilithium.cpp:526`) builds its leaf with `GetScriptForDilithiumThreshold`, which gives the accumulator form:

```
OP_0 {OP_TOALTSTACK <pk> OP_CHECKSIGDILITHIUM OP_FROMALTSTACK OP_ADD}... <m> OP_GREATERTHANOREQUAL
```

`Solver()` only returns `DILITHIUM_MULTISIG` through `MatchDilithiumMultisig`, and that matcher requires a trailing `OP_CHECKMULTISIGDILITHIUM`. The accumulator leaf is classified as nonstandard, so `IsDilithiumLeafSpendable` hits `default: return false`. The RPC calls `CreateP2MR(*wallet, {leaf}, label)` without the opt-out, so every call fails with "P2MR tree contains a leaf the wallet cannot safely spend". That includes the 2-of-3 flow in `test/functional/wallet_dilithium_psbt_multisig.py`.

Recognising the template in Solver would not be enough on its own. In the normal flow each cosigner holds one key, which is below the threshold, so the `available >= required` check would still reject it. When registering a shared multisig, the check should be "the wallet holds at least one participating key" (for example via `ParseP2MRDilithiumLeaf`), not "the wallet can spend alone".

The same problem carries into the stacked #183.

_Checked by Codex and Opus._

BarneyChambers:
CreateP2MR no longer requires this wallet to be able to spend the leaf alone.

createdilithiummultisig builds a threshold-accumulator leaf via GetScriptForDilithiumThreshold. Solver never classified that as DILITHIUM_MULTISIG, and even if it had, each cosigner holds one key so available >= required still failed. The allowlist now uses ParseP2MRDilithiumLeaf and accepts a recognised Dilithium template when this wallet holds at least one of the keys. A 2-of-3 with one local key creates; a tree with none of our keys still fails.

Unit test: create_allows_threshold_accumulator_with_one_cosigner_key.

BarneyChambers:
Pushed a follow-up for the leftover RPC mapping:

- Error text matches the guard: the wallet does not participate in that leaf.
- `sendtop2mr` defaults to `-4`. `-6` only when the lowered message is a coin-selection failure (covers `CreateTransaction`'s `'Insufficient funds'`).
- Functional test: OP_TRUE + `allow_trivial_leaves` + 1e6 BTC expects `-6`.

augchan42:
## Follow-up review: allowlist fix confirmed; two low items

**Fixed:** `createdilithiummultisig` works again. `CreateP2MR` now requires the wallet to hold at least one key in each leaf (`IsDilithiumLeafParticipating`), and it recognises the accumulator template through `ParseP2MRDilithiumLeaf`. It is still checked per leaf: a tree that mixes a wallet-key leaf with an `OP_TRUE` leaf is rejected. `create_allows_threshold_accumulator_with_one_cosigner_key` covers the fix.

**Low: `sendtop2mr` returns -4 for two coin-selection failures** (`src/wallet/rpc/p2mr.cpp:131-136`). Only messages containing "insufficient" or "no spendable" map to -6. These two coin-selection errors contain neither word, so they return -4:
- "The inputs size exceeds the maximum weight…" (`coinselection.cpp:24`). This case is realistic with Dilithium inputs of about 4,400 WU against the 400k standard cap.
- "Unconfirmed UTXOs are available, but spending them creates a chain…" (`spend.cpp:912`).

"no spendable" is only ever produced by `createp2mrspend`, not `FundP2MR`. Before this change every `CreateTransaction` failure returned -6, as Core's `SendMoney` does. Suggest -6 for any `CreateTransaction` failure.

**Note:** an accumulator leaf where the wallet holds at least *m* keys is now `ISMINE_SPENDABLE` rather than watch-only, so those coins enter normal coin selection. From reading the code, signing (`sign.cpp:447`) and dummy sizing (`spend.cpp:169-176`) both handle the accumulator leaf, so this looks fine; I didn't run it. Worth a line in the commit message.

**Test gaps (low):**
- No test for a mixed tree (wallet-key leaf plus `OP_TRUE` leaf), so nothing pins the per-leaf rule.
- No `CHECKMULTISIGDILITHIUM`-template case.
- No -4 `CreateTransaction` failure case.

_Checked by Codex and Opus._

PR #182, opened 2026-08-25
wallet: implement importdilithiumkey rescan

BarneyChambers:
## Summary
- `importdilithiumkey` advertised `rescan=true` and then no-oped (`// TODO`).
- Import now calls `RescanFromTime` the same way `importprivkey` does, after dropping `cs_wallet`.
- Pruned chain or an already-running rescan fail closed.

Closes E002 from the BTQ-Core final report.

## Test plan
- [ ] `test/functional/wallet_dilithium_import_restart.py --legacy-wallet`
- [ ] Confirm `rescan=false` misses coins that already exist, then `rescanblockchain` finds them
- [ ] Confirm `rescan=true` sees those coins without a second RPC

### DIFF
--- a/src/wallet/rpc/dilithium.cpp
+++ b/src/wallet/rpc/dilithium.cpp
@@ -18,6 +18,7 @@
 #include <script/sign.h>
 #include <script/interpreter.h>
 #include <core_io.h>
+#include <rpc/protocol.h>
 #include <rpc/rawtransaction_util.h>
 #include <util/message.h>
 #include <util/strencodings.h>
@@ -26,6 +27,20 @@
 #include <univalue.h>
 
 namespace wallet {
+
+namespace {
+static const int64_t TIMESTAMP_MIN = 0;
+
+void RescanWallet(CWallet& wallet, const WalletRescanReserver& reserver, int64_t time_begin = TIMESTAMP_MIN, bool update = true)
+{
+    int64_t scanned_time = wallet.RescanFromTime(time_begin, reserver, update);
+    if (wallet.IsAbortingRescan()) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
+    } else if (scanned_time > time_begin) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Rescan was unable to fully rescan the blockchain. Some transactions may be missing.");
+    }
+}
+} // namespace
 
 RPCHelpMan getnewdilithiumaddress()
 {
@@ -91,11 +106,14 @@ RPCHelpMan importdilithiumkey()
     return RPCHelpMan{"importdilithiumkey",
         "\nAdds a Dilithium private key (as returned by dumpprivkey) to your wallet and\n"
         "creates a matching single-leaf P2MR receive destination for it.\n"
-        "If 'label' is specified, it is assigned to the new address.\n",
+        "If 'label' is specified, it is assigned to the new address.\n"
+        "\nNote: This call can take over an hour to complete if rescan is true, during that time, other rpc calls\n"
+        "may report that the imported key exists but related transactions are still missing, leading to temporarily incorrect/bogus balances and unspent outputs until rescan completes.\n"
+        "Note: Use \"getwalletinfo\" to query the scanning progress.\n",
         {
             {"privkey", RPCArg::Type::STR, RPCArg::Optional::NO, "The Dilithium private key (see dumpprivkey)"},
             {"label", RPCArg::Type::STR, RPCArg::Optional::OMITTED, "An optional label"},
-            {"rescan", RPCArg::Type::BOOL, RPCArg::Default{true}, "Rescan the wallet for transactions"},
+            {"rescan", RPCArg::Type::BOOL, RPCArg::Default{true}, "Scan the chain and mempool for wallet transactions."},
         },
         RPCResult{
             RPCResult::Type::OBJ, "", "",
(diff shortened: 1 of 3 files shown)


augchan42:
## Independent review — descriptor indexed rescans can miss funds

The rescan wiring is right — reserver, pruned-chain guard, abort and partial-scan handling all look correct. There is one gap underneath it that this PR inherits rather than causes.

### High

`importdilithiumkey` rescans from genesis, but on a descriptor wallet running `-blockfilterindex` the scan takes the fast path. `src/wallet/wallet.cpp:1858` builds `FastWalletRescanFilter` from `GetAllScriptPubKeyMans()` only. P2MR destinations never enter a `ScriptPubKeyMan`: `CreateP2MR` persists them through `SetAddressBook` and `SetP2MRMetadata`, and `IsMine` reaches them at `wallet.cpp:1596` via `GetTrackedP2MRScriptIsMine`.

So the filter cannot contain the imported P2MR scriptPubKey. A historical block whose only relevant output is that P2MR output does not match, the scan records it as scanned, and the funds are missed. The rescan reports success.

Two ways out: add tracked P2MR scriptPubKeys to the fast filter, or skip the optimization when P2MR metadata takes part in a scan. The first is better if the filter can be kept current as entries are added.

`test/functional/wallet_dilithium_import_restart.py` uses `descriptors=False` throughout, so it cannot see this. Please add a case with a descriptor wallet, `-blockfilterindex=1`, and funding confirmed before the import.

The same filter gap applies to `importp2mr` on #183. One fix should cover both — let us settle the approach here and I will not repeat it there.


---

augchan42:
**Review findings**

The rescan wiring in this PR looks correct. It does a full rescan from genesis, takes locks in the same order as upstream `importprivkey`, handles pruned nodes the same way, and reports failures correctly. The P2MR addition to the fast-rescan filter fixes a real gap.

**Pre-existing, High: `importdilithiumkey` can abort the node**

This PR doesn't introduce it, but it moves this code, so it could be fixed here. Neither the RPC nor `StoreDilithiumKeyInWallet` (`src/wallet/p2mr.cpp:605`) checks `WALLET_FLAG_DISABLE_PRIVATE_KEYS`. On a legacy wallet this reaches:

```cpp
// src/wallet/scriptpubkeyman.cpp:992, LegacyScriptPubKeyMan::AddDilithiumKeyPubKeyWithDB
assert(!m_storage.IsWalletFlagSet(WALLET_FLAG_DISABLE_PRIVATE_KEYS));
```

Repro (by code reading, not run): `createwallet w true false "" false false` (legacy, private keys disabled), then `importdilithiumkey <key>`, and the node aborts. Upstream `importprivkey` rejects this case up front with "Cannot import private keys to a wallet with private keys disabled". The same guard is needed here.

**Low / unverified**
- The import doesn't lower the wallet birth time (upstream imports with timestamp 1 via `UpdateTimeFirstKey`), so a later load-time rescan might start after the key's history.
- With `rescan=false` there is no `MarkDirty()`, so cached credit on existing wallet txs paying the new script may stay stale until restart.
- `AddP2MRScriptPubKeys` copies every P2MR metadata blob on every block during a fast rescan. That slows rescans but doesn't affect correctness.

_Checked by Opus._

BarneyChambers:
importdilithiumkey now refuses disable_private_keys wallets, marks dirty, and sets birth time 1 on both wallet types.

The High abort is closed: both the RPC and ImportDilithiumKeyAsP2MR check WALLET_FLAG_DISABLE_PRIVATE_KEYS before StoreDilithiumKeyInWallet, so the legacy assert is not reached. rescan=false now MarkDirty()s so cached credits recompute.

Birth time 1 is UpdateTimeFirstKey on legacy, and on descriptor wallets it persists creation_time 1 on the descriptor that took the key so reload rebuilds m_birth_time. That stops blockConnected from skipping pre-wallet blocks. It does not rewind a stored best-block locator, same as importprivkey. DILITHIUM_KEYMETA load no longer goes through a null legacy manager on descriptor wallets. wallet_dilithium_import_restart.py covers the nokeys abort and unloads/reloads the descriptor import.

BarneyChambers:
Pushed a follow-up so `importdilithiumkey` matches `importprivkey` on the operator-facing bits:

- `EnsureWalletIsUnlocked` before storing the key (`-13` on a locked wallet).
- `LabelFromValue` so `\"*\"` is rejected.
- Help text: new backup required, `rescan=false` needs `rescanblockchain`, passphrase required.

augchan42:
## Follow-up review: earlier items fixed; one new medium

**Fixed:**
- **Private-keys-disabled abort:** now rejected first, before any lock or write, on both wallet types (`dilithium.cpp:138`, `p2mr.cpp:~677`).
- **Birth time:** legacy calls `UpdateTimeFirstKey(1)`; descriptor uses the new lower-only `UpdateTimeFirstKey` plus `FirstKeyTimeChanged`.
- **MarkDirty:** called under `cs_wallet` before the key is added.
- **Order of checks:** unlock and label checks run in the same order as `importprivkey`.

**Medium: the IsMine cache isn't cleared when the key is added.** `CWallet::IsMine` caches results in `m_ismine_cache` (`wallet.cpp:1597-1611`). For tracked P2MR entries the cached value depends on whether the wallet holds the key. `StoreDilithiumKeyInWallet` never clears the cache, and `CreateP2MR` returns early for an existing entry (`p2mr.cpp:711-717`) before `SetP2MRMetadata` would erase it. Every other import path clears the cache (`wallet.cpp:1746/1758/1770`).
- **Repro (by reading, not run):**
  1. Track a P2MR script containing a pubkey the wallet doesn't hold (`getnewp2mraddress` with that leaf, or `createdilithiummultisig`).
  2. Receive funds; the output is cached `ISMINE_WATCH_ONLY`.
  3. `importdilithiumkey` that key.
  4. `MarkDirty` and the rescan recompute with the stale WATCH_ONLY, so the funds aren't spendable until restart.
- **Fix:** clear `m_ismine_cache` after the key is stored.

**Low:**
- **Arbitrary descriptor gets the birth time:** `GetAllScriptPubKeyMans()` is a `std::set` ordered by pointer address, so the key and creation time 1 go to whichever descriptor has the lowest address. On reload, `LoadDilithiumKey` may attach the key to a different one. The wallet birth time is still correct, but an unrelated descriptor then shows timestamp 1 in `listdescriptors`, and a restore via `importdescriptors` rescans it from genesis.
- **Label not updated:** re-importing with a new label keeps the old one (`CreateP2MR` early return). Upstream `importprivkey` updates it.
- **Dead code:** the new `WALLET_FLAG_DESCRIPTORS` branch in the `DILITHIUM_KEYMETA` loader (`walletdb.cpp:932-941`) can't run, because `LoadLegacyWalletRecords` returns early for descriptor wallets.

**Tests:** `wallet_dilithium_import_restart.py` would still pass without the birth-time and MarkDirty fixes.
- The descriptor case imports with rescan=true, so the reload assertion holds anyway.
- The `norescan` case has no transaction paying the script before import, so there's no cached credit to go stale.
- The descriptor-wallet private-keys-disabled case, the locked wallet (-13) and the `*` label are untested.

_Checked by Codex and Opus._

PR #183, opened 2026-08-25
wallet: dump and restore P2MR metadata

BarneyChambers:
## Summary
- `dumpwallet` writes `p2mr <iso-time> <hex(json)>` records from `ListP2MR`.
- `importwallet` restores those records via `CreateP2MR(..., allow_trivial_leaves=true)` so a backup of a test tree still loads.
- Dilithium keys imported from a dump go through `ImportDilithiumKeyAsP2MR`, not a DilithiumPKHash address-book entry.
- Descriptor wallets get `importp2mr` for the same `listp2mr` JSON.

Stacked on the C001 branch because restore reuses `allow_trivial_leaves`. Retarget to master after C001 merges.

Closes D001 from the BTQ-Core final report.

## Test plan
- [ ] `src/wallet/test/wallet_tests` (`dumpwallet_importwallet_roundtrips_p2mr_metadata` and the Dilithium dump path)
- [ ] `test/functional/wallet_p2mr_backup.py --descriptors`
- [ ] Dump a wallet with a custom multi-leaf tree, import into a fresh wallet, spend it


augchan42:
## CI failure is real, and it is this PR's own change

`Build btqd and run address tests` fails on `dilithium_rpc_tests`, not on anything inherited:

```
fatal error: in "dilithium_rpc_tests/dilithium_rpc_commands_registered":
  NonFatalCheckError: Internal bug detected:
  type != Type::ARR && type != Type::OBJ && type != Type::OBJ_NAMED_PARAMS && type != Type::OBJ_USER_KEYS
fatal error: in "dilithium_rpc_tests/dilithium_rpc_command_help": (same)
*** 2 failures are detected in the test module "BTQ Core Test Suite"
```

Master is green at `05b8d924`, and #181 (this PR's base) passes, so the regression is introduced here.

### Cause

`src/rpc/util.h:224` — the four-argument `RPCArg` constructor asserts that the type is **not** a container, because that overload has no `inner` list to describe the container's contents. Container types must use the overload that takes `std::vector<RPCArg>`.

The new `importp2mr` help declares `tree` as `Type::ARR` through the scalar overload:

```cpp
// src/wallet/rpc/p2mr.cpp:364  (new in this PR)
{"tree", RPCArg::Type::ARR, RPCArg::Optional::NO, "P2MR tree leaves in DFS order"},
```

The two pre-existing occurrences of the same argument, at lines 50 and 99, already do it correctly:

```cpp
{"tree", RPCArg::Type::ARR, RPCArg::Optional::NO, "P2MR tree leaves in DFS order", std::vector<RPCArg>{}, RPCArgOptions{}},
```

### Fix

Match the existing form at line 364, ideally describing the leaf element rather than passing an empty vector:

```cpp
{"tree", RPCArg::Type::ARR, RPCArg::Optional::NO, "P2MR tree leaves in DFS order",
    {
        {"leaf", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED, "A tree leaf"},
    }},
```

Note this fires at RPC *registration*, so every RPC help test trips on it — the two failing cases are the symptom, not the scope.

Found during PR triage ahead of a cross-model review of the open queue.

augchan42:
## Independent review — 2 critical, 3 high; backup/restore can silently lose spendability

Cross-model review (Codex, context-blind) against `.claude/code-review-lanes.md`, reviewed as the incremental diff on top of #181. Verified against the code before posting.

Because P2MR spending requires the Merkle tree, metadata that doesn't survive a round trip means **permanently unspendable funds**. The first two are weighted accordingly.

### [critical] A truncated backup imports "successfully" and drops the trees

`src/wallet/rpc/backup.cpp:612`. A dump line is recognised as a P2MR record only if it *already* has three or more fields and valid hex. So `p2mr`, `p2mr <time>`, or `p2mr <time> <truncated-non-hex>` fails that test, falls through to the ordinary key parser, and is **silently ignored**. There is no `# End of dump` footer check and no expected P2MR record count.

Result: a backup truncated mid-write imports every private key, reports success, and omits the Merkle metadata. The user sees a successful restore and discovers the loss only when trying to spend.

**Fix:** dispatch on `vstr[0] == "p2mr"` *before* validating the remaining fields, and hard-fail on any malformed record. Write and verify a footer manifest carrying the P2MR count (ideally a checksum), and parse the whole file before mutating the wallet.

### [critical] Restore ignores the commitments that would catch corruption

`src/wallet/rpc/p2mr.cpp:395`. `listp2mr` exports `address`, `scriptPubKey`, `merkle_root` and `wallet_address` — but `importp2mr` reads only `tree` and `label`, and the legacy importer likewise ignores the dumped address.

So a single altered byte in a leaf script stays valid hex, produces a different Merkle root and a different address, and import reports success. `CreateP2MR` silently derives and stores the wrong destination. Every field needed to detect this was already in the dump and is thrown away.

**Fix:** decode the exported address and 32-byte root strictly, rebuild the tree, and require the derived address, scriptPubKey and Merkle root to match the supplied commitments before writing anything.

### [high] `importp2mr` throws on every invocation

`src/wallet/rpc/p2mr.cpp:364` — the `RPCHelpMan` bug from my earlier comment, independently rediscovered by this review. `CRPCCommand` constructs the `RPCHelpMan` on each call, so the RPC is unusable and every RPC help test fails. This is the cause of the red CI.

### [high] Descriptor import never rescans

`src/wallet/rpc/p2mr.cpp:397` stores the reconstructed destination and returns immediately. A fresh wallet has no matching entries in `mapWallet`, so the restored address reports **no balance** and `createp2mrspend` cannot select the funded output.

Note this means the PR's own functional test (`wallet_p2mr_backup.py`, the immediate spend at line 49) could not pass even after the RPCHelpMan bug is fixed.

**Fix:** derive or accept an earliest `created_at`, check pruned-block availability, and rescan from that time before reporting success. If a no-rescan mode is wanted, make it explicit and document what it costs.

### [high] Legacy import discards the original creation time

`src/wallet/rpc/backup.cpp:752` uses the dumped timestamp for the *current* rescan, but `CreateP2MR` persists `created_at = GetTime()`. Dump a restored wallet again and it carries the import time, so the next restore starts near `T_import - TIMESTAMP_WINDOW` and can miss funding older than that — worst for watch-only or keyless trees, where no key record supplies an earlier timestamp.

**Fix:** add a restore path that persists the validated original creation time, and use it from both importers. Test dump → import → dump → import with funding well older than `TIMESTAMP_WINDOW`.

### [medium] No leaf-count or size limits on import

`src/wallet/rpc/p2mr.cpp:391`. Leaf counts and script sizes are unbounded, and parsing, tree construction, serialization and storage each duplicate the data. A valid depth-20 tree is `2^20 = 1,048,576` leaves; sibling hashes alone are `1,048,576 x 20 x 32` = **640 MiB** before scripts and JSON copies. The dump-file path isn't covered by any RPC request-size limit.

**Fix:** enforce per-tree leaf-count, per-script byte, total-tree byte and per-call aggregate limits *before* reserving vectors or decoding scripts.

### [medium] Duplicate-tree handling loses labels on ordinary P2MR backups

`src/wallet/rpc/backup.cpp:718`. Dilithium key import synthesises the single-leaf P2MR entry first, and for a normal receive address the key line carries no P2MR label — so that entry lands with an empty label. The later metadata record has the real label, but `CreateP2MR` sees an identical tree, treats it as existing, and returns without updating. The unit test masks this by labelling a `DilithiumPKHash` destination rather than a normal P2MR one.

### [medium] Tests omit every case above

`src/wallet/test/wallet_tests.cpp:422` round-trips `OP_TRUE`, checks address and label, and never verifies a funded balance or a spend. Nothing covers malformed or truncated records, commitment mismatch, oversized trees, duplicate labels, missing keys, or second-generation backups. The descriptor spend test is functional-only — and functional tests don't run in CI (#203) — so none of the critical paths have a CI-visible test.

### Interaction with #181

Worth deciding explicitly: can `importp2mr` restore a leaf that #181's guard would now refuse to *create*? Restoring existing funds should almost certainly be allowed, but that should be a deliberate exemption with a test, not an accident of the two paths not sharing the check.

BarneyChambers:
P2MR outputs carry a Merkle tree of scripts. Spending needs that tree. dumpwallet writes it, importwallet / importp2mr put it back. If the metadata does not survive the round trip, the coins are still in the wallet and cannot be spent.

`importp2mr` declared `tree` as an array using the scalar RPCArg constructor. That overload asserts the type is not a container, because it has no inner list. RPC registration hits that assert, so the command was unusable and every RPC help test failed. CI red is this PR, not master.

A truncated dump, or `p2mr` with a short/non-hex payload, fell through to the ordinary key parser and was ignored. Import reported success, keys were in, trees were not. Same for a flipped leaf: hex still parsed, Merkle root changed, import stored the wrong destination. `listp2mr` already exported address and root. Restore threw them away.

Restore now fails the whole import if the file has no `# End of dump` footer or a p2mr line is malformed. It rebuilds the tree and requires dumped address / merkle_root / scriptPubKey to match before writing. `created_at` is kept, otherwise the next dump carries import time and a later restore can miss funding older than TIMESTAMP_WINDOW. `importp2mr` rescans. Duplicate-tree imports update the label, Dilithium key lines were creating the entry first with an empty one. Leaf count and script size are bounded before we allocate.

augchan42:
## Follow-up review — good progress; six main items remain

Confirmed fixed since the last pass: malformed and truncated records, ordinary string commitment mismatches, the RPCArg assertion, nonempty duplicate labels, and the leaf-count bounds.

Six main items remain, plus two smaller ones at the end. The first is shared with #182 and should be settled there.

**1. Fast-filter gap (shared with #182).** `src/wallet/rpc/p2mr.cpp:430` calls `RescanFromTime`, which takes the block-filter fast path on a descriptor wallet. The filter holds descriptor SPKM scripts only, so imported P2MR outputs can be skipped. Whatever we agree on in #182 applies here unchanged.

**2. The rescan result is discarded.** Line 430 ignores the return value and never checks `IsAbortingRescan()`, then returns success. Your own `RescanWallet` helper in #182 does this correctly — reuse that shape.

**3. Missing `created_at` does not force a genesis scan.** In the collection loop at `src/wallet/rpc/p2mr.cpp:408-417`, an entry with no `created_at` contributes nothing to `nTimeBegin`. Mix one such entry with an entry timestamped last month and the scan starts last month. An unknown creation time has to mean genesis.

**4. The zero sentinel is ambiguous.** The same loop uses `nTimeBegin == 0` to mean "unset", so an explicit `created_at: 0` is overwritten by any later positive timestamp. Use `std::optional<int64_t>` rather than 0.

**5. A duplicate restore keeps the wrong creation time.** `src/wallet/p2mr.cpp:731` resolves to `entry.created_at > 0 ? entry.created_at : ts`, and that whole branch only runs when the label differs. This bites on `importwallet`: the key loop calls `ImportDilithiumKeyAsP2MR` first, which stamps the entry with `GetTime()`, and the later `p2mr` metadata restore then finds a duplicate and keeps the import time. The original timestamp is lost and the next rescan starts too late. A restore carrying an explicit `created_at` should win.

**6. Partial application on a bad entry.** Both import paths mutate entries one at a time. A malformed entry part way through throws, the earlier entries stay installed, and no rescan runs. Validate every entry before installing any.

Two smaller ones, whenever you are next in the file:

- In `importwallet`, `if (vstr.size() < 2) continue;` runs before the `p2mr` branch, so a bare `p2mr` line is dropped silently while `p2mr <time>` throws "malformed". Move the record check ahead of the length guard.
- The commitment checks at `src/wallet/p2mr.cpp:771-779` are each guarded by `isStr()`. A `merkle_root` sent as a number is treated as absent instead of rejected.


---

augchan42:
**Review finding (Medium/Low): `created_at: 0` entries don't widen the `importwallet` rescan**

A `created_at` of 0 means the birth time is unknown. `ListP2MR` falls back to 0 when parsing fails, and `p2mr_tests.cpp` checks that the 0 is kept. `dumpwallet` writes such an entry as `p2mr 1970-01-01T00:00:00Z <hex>`. On import (`src/wallet/rpc/backup.cpp:~620`):

```cpp
const int64_t nTime = ParseISO8601DateTime(vstr[1]);
if (nTime > 0) nTimeBegin = std::min(nTimeBegin, nTime);
```

`nTime` is 0, so `nTimeBegin` stays at the tip's block time and `RescanWallet` starts from the tip. Historical transactions to that P2MR output are not found. The key paths in the same function apply `std::min` unconditionally, so a key with time 0 does trigger a full rescan. P2MR metadata should do the same.

The impact is limited: this only affects entries whose `created_at` is already 0, `importwallet` is legacy-only, and a manual `rescanblockchain` recovers the history. No funds are lost.

Suggested fix: in the `p2mr` branch, treat `nTime <= 0` (or `created_at == 0` in the metadata) as "rescan from genesis" (`nTimeBegin = 0`). Add a case to `wallet_p2mr_backup.py` with `created_at: 0` and an output funded before the import.

The High finding on #181 (`createdilithiummultisig` rejected by the leaf allowlist) also applies to this branch.

_Checked by Codex and Opus._

BarneyChambers:
importwallet treats a p2mr timestamp of 0 as rescan from genesis.

created_at 0 dump lines used if (nTime > 0), so nTimeBegin stayed at the tip and history was skipped. Key records already applied min unconditionally. The p2mr branch now sets nTimeBegin = 0 when nTime <= 0.

wallet_p2mr_backup.py jumps mocktime 16 minutes past TIMESTAMP_WINDOW and imports a p2mr-only dump, so key birth times cannot hide a tip-only rescan. Stacked on #181, so createdilithiummultisig's accumulator leaf is accepted when this wallet holds one key.

BarneyChambers:
Pushed a follow-up for the remaining dump/import rescan holes:

- Restore without `created_at` now persists 0 (genesis rescan) instead of `GetTime()`.
- Duplicate restore merges birth time: 0 wins, otherwise `min(stored, incoming)`, so a later dump cannot shrink the next import window.
- `importp2mr` checks pruned block availability before writing metadata.
- Key-only Dilithium `importwallet` stamps the synthesized P2MR record with the dump key time.

BarneyChambers:
Stacked follow-up so this branch does not regress #181's `sendtop2mr` mapping: default `-4`, `-6` only for coin-selection failures (case-insensitive).

augchan42:
## Follow-up review: created_at 0 fix confirmed; two low items

**Fixed:**
- **Rescan start:** `importwallet` sets `nTimeBegin = 0` for `nTime <= 0` (`backup.cpp:619-624`).
- **Persisting 0:** a restore without `created_at` now persists 0 (`p2mr.cpp:832`).
- **Duplicates:** duplicate records merge as "0 wins, else min".
- **`importp2mr`:** keeps 0 through `std::optional`.
- **Other Sep 9 items fixed:** the fast-rescan filter, the rescan result check, validating every entry before any write, the bare `p2mr` line, and non-string commitments.
- **Test:** `wallet_p2mr_backup.py:85-117` isolates the fix well by jumping past the 15-minute `TIMESTAMP_WINDOW`.

**Low: the batch apply is atomic only against validation errors.** `RestoreP2MR` writes each entry separately, and `CreateP2MR` writes the address book and metadata separately (`p2mr.cpp:743-752`). A database write failure partway through, e.g. a full disk, returns an error after earlier entries are committed. Either wrap the batch in one `WalletBatch` transaction, or say "validated before apply" rather than "atomic".

**Low:**
- **Empty import:** `importp2mr []` leaves `time_begin` unset, `value_or(0)` makes it 0, and the call does a full-chain rescan. Return early on an empty array.
- **Qt progress dialog:** the new throws at `backup.cpp:617/628/635/763` skip `showProgress("", 100)`, so the dialog stays up after a malformed dump in Qt.

**Test gap:** the `nTimeBegin = 0` branch is covered only by the functional test, and that test runs only when BDB is compiled.

_Checked by Codex and Opus._

PR #184, opened 2026-08-25
policy: drop default mempool from 2 GB to 300 MB

BarneyChambers:
## Summary
- `DEFAULT_MAX_MEMPOOL_SIZE_MB` was 2000. With PQ-sized witnesses that is an OOM invitation on commodity hardware.
- Default is now 300 MB, same as Bitcoin Core. Operators who want a fat relay set `-maxmempool`.
- Blocksonly floor is unchanged.

Closes E001 from the BTQ-Core final report.

## Test plan
- [ ] Start a node with no `-maxmempool` and check `getmempoolinfo` `maxmempool` is 300000000
- [ ] `-maxmempool=2000` still works for operators who want the old size
- [ ] Blocksonly path still applies its own floor

### DIFF
--- a/src/kernel/mempool_options.h
+++ b/src/kernel/mempool_options.h
@@ -15,8 +15,13 @@
 
 class CBlockPolicyEstimator;
 
-/** Default for -maxmempool, maximum megabytes of mempool memory usage */
-static constexpr unsigned int DEFAULT_MAX_MEMPOOL_SIZE_MB{2000}; // 2 GB
+/** Default for -maxmempool, maximum megabytes of mempool memory usage.
+ *  300 is Bitcoin Core's default and a RAM-safe default for commodity nodes.
+ *  9c97fc7ab raised this to 2 GB so PQ-sized entries would not thrash the
+ *  rolling min-fee. That size is still available with -maxmempool=2000.
+ *  A smaller default does not skip verification (txs are checked before
+ *  TrimToSize); it only shortens retained backlog. */
+static constexpr unsigned int DEFAULT_MAX_MEMPOOL_SIZE_MB{300};
 /** Default for -maxmempool when blocksonly is set.
  *  AppInitMain requires maxmempool >= descendant_size_vbytes * 40. BTQ raised
  *  DEFAULT_DESCENDANT_SIZE_LIMIT_KVB with MAX_BLOCK_WEIGHT, so that floor is
--- a/src/test/mempool_args_tests.cpp
+++ b/src/test/mempool_args_tests.cpp
@@ -14,6 +14,7 @@
 //! compiled out and unrun on every machine this suite is likely to run on,
 //! which is the same shape of untested guard the cap exists to replace.
 
+#include <kernel/mempool_options.h>
 #include <node/mempool_args.h>
 
 #include <test/util/setup_common.h>
@@ -60,6 +61,13 @@ BOOST_AUTO_TEST_CASE(maxmempool_is_capped_on_32bit)
         BOOST_CHECK_MESSAGE(message.find("32-bit") != std::string::npos,
                             "error does not mention the architecture: " + message);
     }
+}
+
+BOOST_AUTO_TEST_CASE(unconfigured_mempool_default_is_300mb)
+{
+    kernel::MemPoolOptions opts;
+    BOOST_CHECK_EQUAL(DEFAULT_MAX_MEMPOOL_SIZE_MB, 300U);
+    BOOST_CHECK_EQUAL(opts.max_size_bytes, DEFAULT_MAX_MEMPOOL_SIZE_MB * 1'000'000);
 }
 
 BOOST_AUTO_TEST_SUITE_END()


augchan42:
> **Superseded 2026-09-16.** The disposition below ("close") no longer applies. See the Sep 8 and Sep 15 follow-ups on this PR, and #220, which collects the sizing question and asks what hardware a default BTQ node should run on. The per-entry sizing in this comment is still current and is cited from #220.

## Independent review — this change reverts a documented decision, on a rationale that doesn't hold

Cross-model review (Codex, context-blind) against `.claude/code-review-lanes.md`, verified against the code and git history before posting.

### [high] Reverts commit `9c97fc7ab` without addressing why it was made

The 2 GB value was not inherited drift — it was set deliberately:

> `9c97fc7ab` **mempool: increase DEFAULT_MAX_MEMPOOL_SIZE_MB from 300 MiB to 2 GiB to prevent pathological eviction with PQ-sized transactions and packages**
>
> PQ signatures/keys inflate individual transactions from ~100–300 bytes per input to multi-kilobyte ranges per input […] A mempool default of 300 MiB is too small for steady-state relay/mining under increased block caps and package relay. Raising the default to 2 GiB reduces thrashing/eviction and gives the fee estimator a larger, more stable sample.

This PR restores exactly the value that commit removed, for exactly the workload that commit was about, without citing it.

### The stated rationale is backwards

The new comment says:

> Dilithium witnesses make a larger default worse (more expensive-to-verify junk), not better.

Transactions are **fully verified before mempool size limiting** — `TrimToSize()` runs on accepted entries. Lowering the retention cap does not avoid the verification cost the comment cites; it only discards work already paid for. Retention and verification are different budgets.

### Sizing

A minimal 1-in/1-out single-leaf P2MR Dilithium transaction: 2,421-byte signature + 1,316-byte leaf script + 1-byte control block, witness `1 + (3+2421) + (3+1316) + (1+1) = 3,746` bytes; with a 94-byte stripped tx and 2-byte marker/flag, weight is `94 x 15 + 3,842 = 5,252 WU` (~329 vB). With mempool overhead, ~4.9 KB per entry.

| Default | Entries held | Blocks of backlog | Wall clock at 60 s spacing |
|---|---|---|---|
| 2 GB (current) | ~408,000 | ~282 | ~4.7 hours |
| 300 MB (this PR) | ~61,000 | ~42 | **~42 minutes** |

Once full, `TrimToSize()` evicts the lowest-feerate packages and raises the rolling minimum fee by the incremental relay fee, so filling the pool with 1 sat/vB traffic pushes the admission floor to 2 sat/vB and suppresses honest low-fee relay while it decays. That is the pathological eviction `9c97fc7ab` was written to prevent.

### Also

**[low]** No test asserts the default of an unconfigured `MemPoolOptions`. The existing mempool-arg tests only pass 300 as an *explicit* value and would pass unchanged at 2,000 too — so this default can be changed in either direction with nothing going red. Moot if the PR is closed.

### Checked and found fine

The hardcoded `* 40` in `DEFAULT_BLOCKSONLY_MAX_MEMPOOL_SIZE_MB` is **not** a stale `WITNESS_SCALE_FACTOR` assumption — it is upstream's 40x descendant-set factor, predating SegWit. `2000 kvB x 1000 x 40 = 80 MB`, and 300 MB clears that floor by 3.75x. No startup regression.

### Suggested disposition

Close, unless there is a measured backlog target that justifies a lower number. If a smaller default is genuinely wanted for constrained nodes, derive it from a chosen BTQ block-backlog horizon using measured P2MR `DynamicMemoryUsage` rather than copying Core's literal — and say in the commit message why `9c97fc7ab` no longer applies.

BarneyChambers:
Review said close this. 2 GB was not inherited drift. `9c97fc7ab` raised it on purpose so PQ-sized entries would not thrash `TrimToSize` and shove the rolling min-fee up to 2 sat/vB.

The comment I originally wrote was backwards. Txs are fully verified before the mempool size cap runs. Shrinking the pool does not skip Dilithium verify, it only throws away work we already paid for and shortens how much backlog we retain (~40 minutes of 60s blocks at ~5 KB per P2MR entry, vs hours at 2 GB).

I kept 300 MB as the default for machines that cannot hold 2 GB of mempool. The comment now says that, and points at `-maxmempool=2000` for a fat relay. Unit test asserts the unset `MemPoolOptions` default so this cannot be flipped again with nothing going red.

augchan42:
## Follow-up review — nothing further needed

The rationale is accurate now — shrinking the mempool affects retention after verification, not verification itself — and the previous 2 GB decision is recorded. The new test covers the default.

The 300 MB value itself is a policy call and it is mine, not something you need to defend. Nothing outstanding from me here.


---

augchan42:

The head (dca496648) is the one I reviewed on Sep 8. Nothing has changed since.

This is policy, not consensus. It only changes how much a node keeps in its mempool, and nodes with different settings still agree on which blocks are valid.

- The comment is accurate. It says why 9c97fc7ab raised the default and how to get 2 GB back.
- The new unit test checks the unset `MemPoolOptions` default, so the value can't change again without a test failing.
- The manpages and `share/examples/btq.conf` already say 300. They were never regenerated for 2000, so this PR makes the code match them.
- The blocksonly floor is unaffected. 300 MB is above the 80 MB minimum that `AppInitMain` requires.

Merge order: #188 adds a test case at the same place in `src/test/mempool_args_tests.cpp`. Whichever of the two merges second needs a rebase that keeps both cases. #185 also edits `src/kernel/mempool_options.h`, but it merges cleanly with this.

Tested:

- Built on the #212–#215 stack with this PR and #192. `mempool_args_tests` passes, including the new case.
- Full functional suite with `check_known_failures.py`: 304 tests, 257 passed, 38 failed (all 38 on the known-failures list, none new), 9 skipped.

Not for this PR: `doc/release-notes/release-notes-btq-0.1.0.md` announced the raise to 2,000 MB. The next release notes should say the default is back to 300 MB and that `-maxmempool=2000` restores the old size.

---

---
**Update 2026-09-16.** The value itself is now under discussion in #220, which compares 300 MB, 1 GB and 2,000 MB against memory use, retained backlog and the cost of filling the mempool. This PR stays open until the external audit finishes.

PR #185, opened 2026-08-25
policy: cap ancestor and package limits at one block

BarneyChambers:
## Summary
- Default ancestor/descendant size was 2000 kvB and `MAX_PACKAGE_WEIGHT` was 40 MW, larger than an 8 MW block.
- Ancestor and descendant limits are now 500 kvB. `MAX_PACKAGE_WEIGHT` is 8e6.
- Blocksonly mempool-size assert loosened from `>= 80` to `>= 20` so the smaller package envelope still compiles.
- Comments in `txpackage_tests.cpp` and `doc/audit-briefing.md` match the new one-block rule.

This is the package/ancestor slice of #94 / A005. It does not close the rest of the resource envelope.

## Test plan
- [ ] `src/test/test_btq -t txpackage_tests`
- [ ] Submit a package at the old 40 MW cap and confirm it is rejected
- [ ] Submit a package that fits in one 8 MW block and confirm it is accepted

### DIFF
--- a/doc/audit-briefing.md
+++ b/doc/audit-briefing.md
@@ -86,9 +86,9 @@ from a constant that moved sitting next to one that did not.
 | `MAX_STANDARD_P2WSH_SCRIPT_SIZE` | 3,600 | 65,536 | `policy/policy.h` |
 | `MAX_STANDARD_SCRIPTSIG_SIZE` | 1,650 | 15,000 | `policy/policy.h` |
 | `MAX_STANDARD_P2WSH_STACK_ITEM_SIZE` | 80 | 15,000 | `policy/policy.h` |
-| `MAX_PACKAGE_WEIGHT` | 404,000 | 40,000,000 | `policy/packages.h` |
-| `DEFAULT_ANCESTOR_SIZE_LIMIT_KVB` | 101 | 2,000 | `policy/policy.h` |
-| `DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` | 101 | 2,000 | `policy/policy.h` |
+| `MAX_PACKAGE_WEIGHT` | 404,000 | 7,584,000 (one default 7.6 MW template) | `policy/packages.h` |
+| `DEFAULT_ANCESTOR_SIZE_LIMIT_KVB` | 101 | 474 (default template at WSF=16) | `policy/policy.h` |
+| `DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` | 101 | 474 (default template at WSF=16) | `policy/policy.h` |
 
 ### Chain
 
@@ -134,24 +134,19 @@ legacy non-witness Dilithium spend puts its signature in the scriptSig at 16
 weight units per byte, four times the upstream penalty, which caps such a
 transaction at about seven inputs.
 
-### The mempool's size limits are now larger than a block
+### Package and ancestor limits are one block
 
-A block holds 8,000,000 weight units, which at a scale factor of 16 is 500,000
-vbytes. Against that, `DEFAULT_ANCESTOR_SIZE_LIMIT_KVB` and
-`DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` are 2,000 kvB, four times a block, and
-`MAX_PACKAGE_WEIGHT` is 40,000,000, five times `MAX_BLOCK_WEIGHT`. Upstream
-holds all three near a tenth of block capacity, which is what makes an accepted
-package always minable in one block. That ordering is inverted here.
-
-The 40,000,000 looks derived rather than chosen: `policy/packages.h` asserts
+A default template holds 7,600,000 weight units minus a 4000 WU coinbase
+reserve, which at a scale factor of 16 is 474,749 vbytes. `DEFAULT_ANCESTOR_SIZE_LIMIT_KVB` and `DEFAULT_DESCENDANT_SIZE_LIMIT_KVB`
+are 474 kvB, and `MAX_PACKAGE_WEIGHT` is 7,584,000, so an accepted package fits
+the default BlockAssembler. `policy/packages.h` asserts
 `MAX_PACKAGE_WEIGHT >= DEFAULT_ANCESTOR_SIZE_LIMIT_KVB * WITNESS_SCALE_FACTOR *
-1000`, so the 2,000 kvB ancestor limit forced it. The ancestor limit is
-therefore the value to ask about.
+1000`; those values now match.
 
-Separately, `MAX_PACKAGE_COUNT` is still 25, and twenty-five transactions at the
-standard ceiling of 400,000 weight units total 10,000,000 — a quarter of the
-package weight limit. In ordinary operation the count limit always binds first
-and the weight limit is close to dead code.
+`MAX_PACKAGE_COUNT` is still 25. Twenty-five transactions at the standard
+ceiling of 400,000 weight units total 10,000,000, so the weight limit binds
+before the count limit for max-size standard txs. That is the intended envelope
+for issue #94's package/ancestor question.
 
 ### Per-input weight changes what a wallet can spend
 
--- a/doc/audit-briefing.md
+++ b/doc/audit-briefing.md
@@ -137,9 +137,10 @@ transaction at about seven inputs.
 ### Package and ancestor limits are one block
 
 A default template holds 7,600,000 weight units minus a 4000 WU coinbase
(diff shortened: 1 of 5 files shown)


augchan42:
## Independent review — the one-block premise is off by ~5%

### [medium] "Mineable in a single block" does not hold at the new limit

The new comment in `src/policy/policy.h:67` states:

> 500 kvB is one 8 MW block at WSF=16, so an accepted ancestor set is mineable in a single block.

The weight arithmetic is right — `500 x 1,000 x 16 = 8,000,000 WU` = `MAX_BLOCK_WEIGHT`. But no block template is ever built to `MAX_BLOCK_WEIGHT`:

- `DEFAULT_BLOCK_MAX_WEIGHT = 7,600,000` — `src/policy/policy.h:24` (7.6 MW soft cap)
- `nBlockWeight = 4000` reserved at start — `src/node/miner.cpp:98`
- `TestPackage` rejects on `nBlockWeight + WITNESS_SCALE_FACTOR * packageSize >= nBlockMaxWeight` — `src/node/miner.cpp:204`

Solving `4000 + 16 * vsize < 7,600,000` gives **vsize <= 474,749 vB**.

So this node's own `BlockAssembler` cannot mine an ancestor set between 474,750 and 500,000 vB — roughly the top 5% of the range the new policy admits. The invariant the change is justified by fails exactly where it matters.

**Fix, whichever you prefer:**
- Set both size defaults to 474 kvB and `MAX_PACKAGE_WEIGHT` to `474 x 1,000 x 16 = 7,584,000`, and add a template-boundary test; or
- Keep 500 kvB and drop the mineability claim, documenting that packages above the template budget span blocks.

### [low] The changed test pins none of this

`src/test/txpackage_tests.cpp` changes **only comments** in this diff. The oversized-package loop derives its input from `MAX_PACKAGE_WEIGHT`, so it passes at both 40,000,000 and 8,000,000 — reverting the policy change fails no test. Nothing covers the 500 kvB ancestor/descendant defaults, the 20 MB blocks-only default, or exact-limit versus limit-plus-one behaviour. With no functional tests in CI (#203), a C++ boundary assertion is worth more here than a functional one.

### Verified correct

All the derived arithmetic checks out:

- `MAX_BLOCK_WEIGHT / WITNESS_SCALE_FACTOR = 8,000,000 / 16 = 500,000 vB` ✓
- Blocks-only floor: `ceil(500 x 1,000 x 40 / 1,000,000) = 20 MB`, and `static_assert(... >= 20)` holds ✓
- `MAX_PACKAGE_WEIGHT >= MAX_STANDARD_TX_WEIGHT`: `8,000,000 >= 400,000` ✓
- `MAX_PACKAGE_WEIGHT >= DEFAULT_ANCESTOR_SIZE_LIMIT_KVB * 16 * 1000`: `8,000,000 >= 8,000,000` ✓ (exact equality — worth knowing there is zero margin)
- Same for the descendant assertion ✓

This is policy-only; it cannot change block validity. Nodes that upgrade will drop persisted mempool entries exceeding the new chain limits on reload, which is expected but worth a release note.

BarneyChambers:
Ancestor and descendant size limits are how big a mempool chain of unconfirmed txs we will accept. They were 2000 kvB, larger than an 8 MW block, so a package could be unmineable in one go.

This PR cut them to 500 kvB and said that is one block at WSF=16. The arithmetic for the hard cap is right: 500 * 1000 * 16 = 8,000,000 WU.

The miner never fills to that. Default template is 7.6 MW and it reserves 4000 WU for the coinbase. `TestPackage` rejects when `4000 + 16 * vsize >= 7,600,000`, so vsize > 474,749 never makes a block here. The top ~5% of the 500 kvB range would sit in the mempool and not be selected.

Limits are 474 kvB and `MAX_PACKAGE_WEIGHT` is 7,584,000. There is a unit test so those numbers cannot drift apart.

augchan42:
## Follow-up review — narrow the claim, and one arithmetic slip

The weight arithmetic checks out: 474 kvB x 16 x 1000 = 7,584,000 matches `MAX_PACKAGE_WEIGHT`, and the 19 MB blocksonly floor follows. Two fixes left in `doc/audit-briefing.md`.

The section claims an accepted package "fits the default BlockAssembler". It fits the default **weight** budget. `src/node/miner.cpp:207` checks `MAX_BLOCK_SIGOPS_COST` separately, so a package inside the weight budget can still be turned away on sigops. Please narrow the claim.

Same paragraph gives the template as "474,749 vbytes". 7,600,000 minus the 4,000 WU coinbase reserve is 7,596,000, which at WSF 16 is 474,750.


---

augchan42:
**Review finding (Low): `mempool_limit.py` still expects the old 80 MB floor**

Lowering `DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` to 474 makes the `-maxmempool` floor `474,000 × 40`, which rounds up to 19 MB (`src/init.cpp:1489-1491`). `test/functional/mempool_limit.py:371-375` still asserts `"Error: -maxmempool must be at least 80 MB"`, and the comment above it still mentions 80 MB.

This is currently hidden: the same test starts its node with `-maxmempool=5` (line 33), which is below both the old and new floors, so it already fails at startup on master. Fixing that will expose this assertion.

The rest of the arithmetic checks out:
- 474 kvB × 16 plus the 4000 coinbase reserve fits in `DEFAULT_BLOCK_MAX_WEIGHT`.
- The `static_assert`s hold.
- The CPFP carve-out and the option parsing are consistent.

Nits:
- `src/test/txpackage_tests.cpp:65` says "8 MW"; the ceiling is 7.584 MW.
- `doc/policy/packages.md` still says `MAX_PACKAGE_WEIGHT=404000`. That was already wrong before this PR.

_Checked by Opus._

BarneyChambers:
fill_mempool now actually fills the 5 MB pool.

The 101 kvB ancestor/descendant args keep the floor at 5 MB so eviction can fire, and the 19 MB default-floor check still uses a fresh extra_args list so 18 MB is rejected against 474 kvB * 40. That was not enough on its own: the OP_RETURN fill txs weigh ~1.08 MW at WSF 16, over MAX_STANDARD_TX_WEIGHT, so they died on tx-size before TrimToSize. extra_args now includes -acceptnonstdtxn=1; assert_start_raises_init_error(['-maxmempool=18']) still does not inherit that or the 101 kvB limits.

BarneyChambers:
Merged current master (conflict was `mempool_limit.py`: master's 80 MB floor vs this PR's 19 MB).

`mempool_limit.py` fee math used Core's `weight // 4`. At WSF 16 that made the immediate-eviction package rank *above* the worst mempool entry, and the CPFP carve-out pad was 2501 vB instead of 10001. Fees and pads are in vbytes now. The mid-package parents stay at 25 kvB so three of them fit under the test's 101 kvB ancestor cap. `mempool_limit.py` is off `known_failures.txt`.

augchan42:
## Follow-up review: floor fixed, but `mempool_limit.py` now fails in CI

**Fixed:**
- **Floor assertion:** now 19 MB (`mempool_limit.py:402-405`).
- **101 kvB args:** they keep the floor at 4.04 MB, under `-maxmempool=5`, so the node starts and eviction fires.
- **Fee math:** `vsize = ceil(w/16)` is correct.
- **Carve-out:** 160,001 WU is 10,001 vB, just over the 10,000 vB `EXTRA_DESCENDANT_TX_SIZE_LIMIT`, which is correct.
- **Nits:** the `txpackage_tests.cpp` and `packages.md` wording is fixed.

**High: `mempool_limit.py` fails.** The functional job on 30af6f5a2 (job 106303017759) reports:
```
AssertionError: 100000 <= 135850
mempool_limit.py failed and is not on the known-failures list.   (91 known, 1 new)
```
- **Cause:** line 386 compares `target_vsize * 2` against `maxmempool - bytes`. `bytes` is the sum of vsizes, while trimming works on memory usage, so the leftover measured in vB is about 136k, not the "<65 kvB" that the comments at lines 198 and 383 say.
- **No size fixes it:** under `-limitdescendantsize=101`, parent plus child can't exceed 101 kvB, so no target size can pass this check.
- **Line 203 too:** `parent_vsize * 3` against the same leftover will likely fail next in `test_mid_package_eviction`. The run doesn't get that far.
- **Fix:** compare against `getmempoolinfo()["usage"]`, or drop the assertion. The later "mempool full" check already proves eviction happened.

**Low:** `-acceptnonstdtxn=1` (line 50) covers more than the fill transactions its comment names. The big parents (about 400k WU; anything over fails `tx-size`) and the 800k-WU pair at line 384 also rely on it. The eviction logic under test doesn't depend on standardness, but either document this or size those transactions below 25,000 vB.

_Checked by Codex and Opus._

Issue #144, opened 2026-08-04
launch: reset testnet from a new genesis so it runs mainnet's consensus config


augchan42:
> **Filed 2026-08-04.** Scheduled after the external audit. Spec moved here from Linear BTQ-111 so it lives with the code rather than in a closed ticket.
>
> _Revisions to this issue go at the top, newest first._

Testnet will be restarted from a new genesis so that it runs the consensus configuration mainnet actually launches with. This issue carries the parameter work, so that whoever performs the reset is reading it next to the code rather than reconstructing it.

**Sequencing: after the external audit.** The audit pins a revision; resetting the chain underneath it mid-engagement is not worth the disruption. Filing now so the plan exists in the repo before auditors ask why testnet's consensus config differs from mainnet's.

## Why reset rather than activate on the existing chain

Testnet has `nDilithiumP2MRHeight = std::numeric_limits<int>::max()`; mainnet ships it at `1`. So P2MR-only is enforced on testnet as *policy* — non-standard to relay, refused by the wallet — but not as *consensus*. Legacy Dilithium spends remain consensus-valid there.

The consequence is that **the P2MR-only consensus enforcement path that mainnet launches with has never validated a live chain.** A bug in it cannot be caught by testnet as currently configured.

The alternative — activate at a future height on the existing chain — was conditional on first moving any value worth keeping. #111 measured what that would cost and the condition cannot be met:

> 719,045 BTQ across 141,888 outputs, against a total UTXO value of 956,445 BTQ — about **75% of the chain**. All of it legacy base58 Dilithium P2PKH, because testnet paid every coinbase to one of those until the miners moved to P2MR at height 159,436.
>
> It spans heights 1 to 159,436 fairly evenly, and the largest 1,000 outputs hold only 2% of the value. This is not a handful of pool payout addresses that can be swept before a flag day. It is the chain's entire mining history, spread across everyone who has ever mined it.

Resetting also lets `nLWMAHeight` go to `1`, so testnet exercises mainnet's launch difficulty behaviour — LWMA bootstrapping from genesis with no preceding legacy-retarget history. Nothing has ever run that.

## Parameters

Six consensus params differ between mainnet and testnet in `src/kernel/chainparams.cpp`. They split cleanly.

### Align to mainnet

| Param | Mainnet | Testnet today | Action |
| --- | --- | --- | --- |
| `nDilithiumP2MRHeight` | 1 | `int::max` | set to 1 |
[...]

PR #186, opened 2026-08-25
consensus: keep DILITHIUM_SIGOP_COST=50 pending input benches

BarneyChambers:
## Summary
- `DILITHIUM_SIGOP_COST` stays 50. Changing it forks live testnet, and 50 already over-weights Dilithium if the real ratio is closer to 10.
- `src/script/script.h` and `src/bench/dilithium_verify.cpp` now say how to decide later: run `./src/bench/bench_btq -filter='(P2MRDilithium|P2WPKHECDSA)Input'` and take `ceil(ratio * 1.5)`.
- Keep 50 if that number is <= 50. Raise only if the input-level ratio is well above 50, and pair a raise with a testnet reset. Do not lower 50 without a written DoS argument.

This documents B001 rather than forking the constant. A real constant change is a companion to #144.

## Test plan
- [ ] Read the new comments in `src/script/script.h` and the bench file
- [ ] On release hardware, run `./src/bench/bench_btq -filter='(P2MRDilithium|P2WPKHECDSA)Input'` and record ns/op
- [ ] If `ceil(ratio * 1.5) > 50`, open a follow-up that raises the constant with a testnet reset

### DIFF
--- a/src/bench/dilithium_verify.cpp
+++ b/src/bench/dilithium_verify.cpp
@@ -21,20 +21,22 @@
 // Signature verification cost, Dilithium against ECDSA, measured on the same
 // hardware over the same input.
 //
-// This exists because two consensus constants encode a ratio between the two
-// that has never been measured. DILITHIUM_SIGOP_COST (src/script/script.h)
-// charges a Dilithium check 50 sigops, asserting that one Dilithium
-// verification costs about fifty ECDSA verifications. The reasoning recorded
-// when that was chosen cited a figure closer to 10. Both cannot be right, and
-// the constant bounds how many post-quantum signatures a block can carry:
-// MAX_BLOCK_SIGOPS_COST / WITNESS_SCALE_FACTOR legacy sigops, divided by
-// whatever a Dilithium op is charged.
+// This exists because DILITHIUM_SIGOP_COST (src/script/script.h) charges a
+// Dilithium check 50 sigops. 50 is a provisional launch value: over-weight
+// wastes a bit of block space, under-weight is CPU DoS. The primitive pair
+// below overstates the ratio because script/Merkle overhead is shared.
+//
+// Use the input-level pair (P2MRDilithiumInput / P2WPKHECDSAInput). Compute
+// required = ceil(ratio * 1.5). Keep 50 iff required <= 50; otherwise propose
+// `required` with a testnet reset. The bench is one-input VerifyScript, not
+// UTXO lookup or block-level scheduling.
 //
 // Run both and divide the reported ns/op:
 //
 //     ./src/bench/bench_btq -filter='(Dilithium|ECDSA)Verify'
+//     ./src/bench/bench_btq -filter='(P2MRDilithium|P2WPKHECDSA)Input'
 //
-// The ratio is what DILITHIUM_SIGOP_COST should approximate. Verification is
+// The input-level ratio is what DILITHIUM_SIGOP_COST should approximate. Verification is
 // the operation to measure rather than signing, because validation cost is what
 // the sigop budget exists to bound -- every node verifies every signature in
 // every block, while signing happens once on the spending wallet.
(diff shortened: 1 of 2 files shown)


augchan42:
## Independent review — 2 findings, both in the guidance text

Cross-model review (Codex, context-blind) against `.claude/code-review-lanes.md`. Confirmed first: this diff changes **comments only** — `DILITHIUM_SIGOP_COST` stays 50, and the `P2MRDilithiumInput` / `P2WPKHECDSAInput` benchmarks it points at already existed at the base revision (`src/bench/dilithium_verify.cpp:116,206`, registered at 271–272). So the findings are about whether the guidance is followable, which for a consensus constant is the whole point of the PR.

### [medium] The keep/raise rule is self-contradictory

`src/script/script.h:72` says:

> Keep 50 if `ceil(ratio * 1.5) <= 50`.

which means keep only when ratio ≤ 33.33. Two lines later:

> Raise it only if the input-level ratio is well above 50.

A measured ratio between ~33.4 and 50 satisfies **neither** rule. Worked example: ratio 40 gives `ceil(40 * 1.5) = 60 > 50`, so the first sentence requires a raise while the second discourages one. Keeping 50 there would also violate the 1.5x safety margin the same comment establishes.

**Fix:** state one rule and delete the other, e.g. *"Compute `required = ceil(ratio * 1.5)`. Retain 50 iff `required <= 50`; otherwise propose `required` with an activation/reset plan."* Drop "well above 50" or replace it with the actual number.

### [low] "Conservative" is asserted, not measured

`src/bench/dilithium_verify.cpp:25` calls 50 "the conservative launch value" while the same comment says the input-level ratio still needs measuring, and the PR adds no results. By the comment's own rule, 50 is conservative only if the measured ratio is ≤ 33.3 — which is precisely what is unknown.

**Fix:** call it *provisional* pending recorded measurements, or land the numbers from a supported build/hardware config alongside it.

### Useful result from the review — the constant is doing real work

Worth recording in the comment, since it is the DoS argument the text says is missing:

| Bound | Calculation | Max Dilithium inputs per block |
|---|---|---|
| Sigops | `80,000 / 50` | **1,600** |
| Weight | `8,000,000 / 4,402` | ~1,817 |

where a benchmark-shaped P2MR input is `656` (base `41 x 16`) + `3,746` (witness: `1 + 2,424 + 1,319 + 2`) = **4,402 WU**. At the sigop ceiling, inputs consume `1,600 x 4,402 = 7,043,200 WU`, leaving room for overhead.

**Sigops bind before weight**, so `DILITHIUM_SIGOP_COST` is genuinely throttling input count rather than being slack — which supports keeping it, and is a stronger argument than the text currently makes. The CPU safety margin remains unproven either way.

Note the existing benchmark measures one-input `VerifyScript` (witness/script processing, P2MR control/root handling, sighash assembly, verification). It excludes UTXO lookup, precomputation, sigop counting, and block-level scheduling — a fair script-input ratio, not full block-validation cost. Worth saying so where the comment tells a future reader to trust the ratio.

BarneyChambers:
Comments only, `DILITHIUM_SIGOP_COST` stays 50. Changing it forks live testnet.

The keep/raise text disagreed with itself. "Keep 50 if ceil(ratio * 1.5) <= 50" means raise when the measured ratio is above ~33. Two lines later, raise only if the ratio is well above 50. A measured 40 hit both rules.

One rule now: compute `required = ceil(ratio * 1.5)`, keep 50 iff that is <= 50, otherwise propose `required` with a testnet reset. 50 is called provisional until the input-level bench is recorded.

Worth having in the comment: at 50, sigops bind first. 80000/50 = 1600 Dilithium inputs per block vs ~1817 by weight at ~4402 WU per 1-in P2MR input. The constant is actually throttling, not slack.

augchan42 (approved):

augchan42:
Approved and ready to merge whenever you want to land it. Nothing further from my side.

PR #187, opened 2026-09-01
p2p: enable BIP324 v2 transport by default

BarneyChambers:
## Summary
- Default `-v2transport` on (Bitcoin Core 27.0 #29347).
- `-connect`, `-addnode`, and `-seednode` follow that flag. V1 fallback on reconnect is unchanged.
- `addnode` with no `v2transport` arg uses the node's setting. The functional-test helper now passes the flag explicitly so node-to-node tests stay v1 unless `--v2transport` is set.

## Test plan
- [ ] `p2p_v2_transport.py`
- [ ] Two regtest nodes with no flags: `getpeerinfo` shows `v2` on the outbound
- [ ] `-v2transport=0` still starts and connects v1
- [ ] `connect_nodes` without `--v2transport` still uses v1

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1866,7 +1866,7 @@ bool CConnman::AddConnection(const std::string& address, ConnectionType conn_typ
     CSemaphoreGrant grant(*semOutbound, true);
     if (!grant) return false;
 
-    OpenNetworkConnection(CAddress(), false, std::move(grant), address.c_str(), conn_type, /*use_v2transport=*/false);
+    OpenNetworkConnection(CAddress(), false, std::move(grant), address.c_str(), conn_type, /*use_v2transport=*/bool(GetLocalServices() & NODE_P2P_V2));
     return true;
 }
 
@@ -2325,7 +2325,7 @@ void CConnman::ProcessAddrFetch()
     CAddress addr;
     CSemaphoreGrant grant(*semOutbound, /*fTry=*/true);
     if (grant) {
-        OpenNetworkConnection(addr, false, std::move(grant), strDest.c_str(), ConnectionType::ADDR_FETCH, /*use_v2transport=*/false);
+        OpenNetworkConnection(addr, false, std::move(grant), strDest.c_str(), ConnectionType::ADDR_FETCH, /*use_v2transport=*/bool(GetLocalServices() & NODE_P2P_V2));
     }
 }
 
@@ -2428,7 +2428,7 @@ void CConnman::ThreadOpenConnections(const std::vector<std::string> connect)
             for (const std::string& strAddr : connect)
             {
                 CAddress addr(CService(), NODE_NONE);
-                OpenNetworkConnection(addr, false, {}, strAddr.c_str(), ConnectionType::MANUAL, /*use_v2transport=*/false);
+                OpenNetworkConnection(addr, false, {}, strAddr.c_str(), ConnectionType::MANUAL, /*use_v2transport=*/bool(GetLocalServices() & NODE_P2P_V2));
                 for (int i = 0; i < 10 && i < nLoop; i++)
                 {
                     if (!interruptNet.sleep_for(std::chrono::milliseconds(500)))
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1833,7 +1833,7 @@ void CConnman::CreateNodeFromAcceptedSocket(std::unique_ptr<Sock>&& sock,
     RandAddEvent((uint32_t)id);
 }
 
-bool CConnman::AddConnection(const std::string& address, ConnectionType conn_type)
+bool CConnman::AddConnection(const std::string& address, ConnectionType conn_type, bool use_v2transport)
 {
     AssertLockNotHeld(m_unused_i2p_sessions_mutex);
     std::optional<int> max_connections;
@@ -1866,7 +1866,7 @@ bool CConnman::AddConnection(const std::string& address, ConnectionType conn_typ
     CSemaphoreGrant grant(*semOutbound, true);
     if (!grant) return false;
 
-    OpenNetworkConnection(CAddress(), false, std::move(grant), address.c_str(), conn_type, /*use_v2transport=*/bool(GetLocalServices() & NODE_P2P_V2));
+    OpenNetworkConnection(CAddress(), false, std::move(grant), address.c_str(), conn_type, use_v2transport);
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


BarneyChambers:
Tracking: #195

BarneyChambers:
No review notes. Default `-v2transport` is still on. That is BIP324: encrypted P2P, handshake is still secp256k1 so it is not post-quantum, it stops casual wire observers and DPI that keys on Bitcoin's v1 traffic.

augchan42:
## Independent review — revised; the test-framework gap is wider than I first said

The default flip and the `addnode` default-hint work look right. This will break the functional suite, though, and the fix is larger than I described.

### High

`src/net.cpp:1869` makes `AddConnection()` initiate v2 by default. That call backs the `addconnection` RPC, which is what `TestNode.add_outbound_p2p_connection()` uses. The test framework installs a Python listener on the other end, and the node now opens a v2 handshake into it.

**Correction.** I previously suggested teaching the listener v2 as an alternative fix. I went looking for how much of it would need to change and the answer is all of it: `supports_v2_p2p` does not appear anywhere in `test/functional/test_framework/p2p.py`. `peer_connect_helper` sets `self.magic_bytes` and nothing else — this fork never inherited upstream's v2 test transport. So the listener rejects the handshake on a magic-byte mismatch and burns its one protocol instance, with no reconnection path.

Suggested split:

1. In this PR, keep explicit v1 for the test callers so the suite stays green. `AddConnection()` can take the transport as a parameter rather than reading `GetLocalServices()`.
2. File the framework v2 port separately. It is real work — `EncryptedP2PState`, the v2 handshake, reconnection — and it should not gate the default flip.

Worth a line in the PR body too: `-addnode` entries now follow `-v2transport` (`src/net.h`), which changes behaviour for anyone with a pinned peer list.


---
_Revised 2026-09-09 after several review rounds across Opus and Codex, context-blind against the current head. The correction above supersedes what this comment said earlier._

BarneyChambers:
- `-addnode` entries now follow `-v2transport` (`src/net.h`), which changes behaviour for anyone with a pinned peer list.

augchan42:
**Review findings**

**1. Critical: `addnode` fails whenever `v2transport` is passed**

The `v2transport` fallback changed from `RPCArg::Default{false}` to `RPCArg::DefaultHint{...}`, but a non-null value is still read with `self.Arg<bool>(2)` (`src/rpc/net.cpp:322-324`). `Arg<T>` runs `CheckRequiredOrDefault` before the null check (`src/rpc/util.cpp:730`), and that check requires the arg to be either required or to have a real `RPCArg::Default`. So any explicit `v2transport=true/false` trips `CHECK_NONFATAL` and returns "Internal bug detected".

This PR also changes `test_framework.py` so that `connect_nodes` always passes `v2transport=<bool>`. That breaks setup for nearly every functional test.

Fix (matches upstream):
```cpp
const bool use_v2transport = self.MaybeArg<bool>(2).value_or(bool(connman.GetLocalServices() & NODE_P2P_V2));
```

**2. Medium: `p2p_node_network_limited.py` asserts stale service flags**

Lines 52-58 and 79 hard-code `NODE_WITNESS | NODE_NETWORK_LIMITED`. With `-v2transport` now on by default, `init.cpp` also advertises `NODE_P2P_V2`, so the version-message `nServices`, `getnetworkinfo` `localservices` and self-advertised addr assertions all fail (0xc08 vs 0x408). Upstream ORs in `NODE_P2P_V2` when v2 is enabled.

**Minor / unverified**
- `doc/man/btqd.1`, `doc/man/btq-qt.1` and `share/examples/btq.conf` still say "Support v2 transport (default: 0)".
- `feature_proxy.py` may become flaky. `onetry` connections through the stub SOCKS5 server can trigger a v1 reconnect through the same proxy and leave a stale command in the queue. This depends on timing and wasn't run.

We found no problems with BIP324 on BTQ's parameters: the HKDF salt uses BTQ's magic bytes, the test vectors were regenerated, and v1 detection uses `MessageStart()`.

_Checked by Opus._

BarneyChambers:
addnode with an explicit v2transport no longer hits CHECK_NONFATAL.

The arg is a DefaultHint, so Arg<bool>(2) always asserted. It now uses MaybeArg, same as upstream: omitted follows local NODE_P2P_V2, true/false are taken as given. connect_nodes can pass v2transport again.

p2p_node_network_limited.py now expects NODE_P2P_V2 and NODE_BTQ_DILITHIUM as well as witness + limited. Manpages and btq.conf say v2 default 1.

The Python v2 handshake port is still a follow-up, not this PR.

BarneyChambers:
Merged current master.

Three new functional failures: `--usecli` `addnode(command=...)` collided with `send_cli(command=...)`; v1 `P2PInterface` peers were getting a v2 handshake. `connect_nodes` is positional. Tests pin `-v2transport=0` on `TestNode.args` so `restart_node(extra_args=...)` keeps it. `DEFAULT_V2_TRANSPORT` is locked in `net_tests`. `doc/bips.md` says on by default.

augchan42:
## Follow-up review: fixes confirmed; two low items

**Fixed:**
- **`addnode`:** uses `self.MaybeArg<bool>(2).value_or(<local NODE_P2P_V2>)` (`rpc/net.cpp:322`), the same as upstream, and the explicit-true-while-disabled error is kept.
- **Service flags:** `p2p_node_network_limited.py` expects WITNESS | NETWORK_LIMITED | P2P_V2 | BTQ_DILITHIUM and passes in CI.
- **Docs:** man pages and `btq.conf` say default 1.
- **`--usecli`:** the collision is fixed by calling `addnode` positionally.
- **CI:** the functional job shows 0 new failures, and `feature_proxy.py` passes.

**Low: stale known-failures entry.** `known_failures.txt:46` still lists `p2p_node_network_limited.py`. CI warns that it passes. While it's listed, a future regression in that test won't turn CI red.

**Low: nothing tests the new default end to end.** The harness pins `-v2transport=0` on ordinary test nodes (`test_framework.py:496`). `net_tests.cpp:39` checks only the constant, and `p2p_v2_transport.py` passes `=1` explicitly. No functional test starts a node without the flag and checks that `NODE_P2P_V2` is in `localservices`, or that `addnode` without the third argument makes a v2 connection. Suggest adding one.

**Nit:** the pin only matches exact strings in the initial `extra_args`. A test that later calls `restart_node(extra_args=...)` without `=0` would come back with v2 on. No current test does this.

_Checked by Codex and Opus._

PR #188, opened 2026-09-01
policy: default full replace-by-fee on

BarneyChambers:
## Summary
- `DEFAULT_MEMPOOL_FULL_RBF` is now true (Bitcoin Core 28.0 #30493).
- `-mempoolfullrbf=0` still restores BIP125 opt-in signaling.
- `feature_rbf.py` pins `-mempoolfullrbf=0` so the signaling tests stay meaningful. `test_fullrbf` still turns it on.

## Test plan
- [ ] `feature_rbf.py`
- [ ] `getmempoolinfo` shows `"fullrbf": true` on a default node
- [ ] A replacement with `SEQUENCE_FINAL` is accepted at default settings

### DIFF
--- a/doc/man/btq-qt.1
+++ b/doc/man/btq-qt.1
@@ -674,7 +674,7 @@ this size or less (default: 83)
 \fB\-mempoolfullrbf\fR
 .IP
 Accept transaction replace\-by\-fee without requiring replaceability
-signaling (default: 0)
+signaling (default: 1)
 .HP
 \fB\-minrelaytxfee=\fR<amt>
 .IP
--- a/doc/man/btqd.1
+++ b/doc/man/btqd.1
@@ -674,7 +674,7 @@ this size or less (default: 83)
 \fB\-mempoolfullrbf\fR
 .IP
 Accept transaction replace\-by\-fee without requiring replaceability
-signaling (default: 0)
+signaling (default: 1)
 .HP
 \fB\-minrelaytxfee=\fR<amt>
 .IP
--- a/doc/policy/mempool-replacements.md
+++ b/doc/policy/mempool-replacements.md
@@ -10,13 +10,13 @@ A transaction ("replacement transaction") may replace its directly conflicting t
 their in-mempool descendants (together, "original transactions") if, in addition to passing all
 other consensus and policy rules, each of the following conditions are met:
 
-1. The directly conflicting transactions all signal replaceability explicitly. A transaction is
-   signaling replaceability if any of its inputs have an nSequence number less than (0xffffffff - 1).
+1. The directly conflicting transactions all signal replaceability explicitly, unless
+   `-mempoolfullrbf` is enabled (the default). A transaction is signaling replaceability if any
+   of its inputs have an nSequence number less than (0xffffffff - 1).
 
    *Rationale*: See [BIP125
    explanation](https://github.com/btq/bips/blob/master/bip-0125.mediawiki#motivation).
-   Use the (`-mempoolfullrbf`) configuration option to allow transaction replacement without enforcement of the
-   opt-in signaling rule.
+   `-mempoolfullrbf=0` restores the opt-in signaling rule.
 
 2. The replacement transaction only include an unconfirmed input if that input was included in
    one of the directly conflicting transactions. An unconfirmed input spends an output from a
--- a/share/examples/btq.conf
+++ b/share/examples/btq.conf
@@ -566,7 +566,7 @@
 #datacarriersize=1
 
 # Accept transaction replace-by-fee without requiring replaceability
-# signaling (default: 0)
+# signaling (default: 1)
 #mempoolfullrbf=1
 
 # Fees (in BTQ/kvB) smaller than this are considered zero fee for
--- a/src/kernel/mempool_options.h
+++ b/src/kernel/mempool_options.h
@@ -27,7 +27,7 @@ static_assert(DEFAULT_BLOCKSONLY_MAX_MEMPOOL_SIZE_MB >= 80);
 /** Default for -mempoolexpiry, expiration time for mempool transactions in hours */
 static constexpr unsigned int DEFAULT_MEMPOOL_EXPIRY_HOURS{336};
 /** Default for -mempoolfullrbf, if the transaction replaceability signaling is ignored */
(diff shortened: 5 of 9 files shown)


BarneyChambers:
Tracking: #196

augchan42:
## Independent review — 3 findings; two functional tests break silently

Cross-model review (Codex, context-blind) against `.claude/code-review-lanes.md`. All three verified against the code before posting. The runtime change itself is correct and policy-only — the gaps are in tests and docs.

### [medium] The new default breaks two functional tests this PR doesn't touch

Both verified present, both asserting the old default:

**`test/functional/mempool_accept.py:154`** expects `txn-mempool-conflict` for an opt-out replacement paying more fee:
```python
tx.vout[0].nValue -= int(4 * fee * COIN)  # Set more fee
self.check_mempool_result(
    result_expected=[{'txid': tx.rehash(), 'allowed': False, 'reject-reason': 'txn-mempool-conflict'}],
```
Under full-RBF the mempool tx pays `2f` and the candidate pays `6f`, so it is now **accepted** and the assertion fails.

**`test/functional/p2p_permissions.py:122`** expects the same reject reason. With full-RBF the tx reaches the replacement fee checks instead and fails with `insufficient fee`.

Upstream updated both in the corresponding commit ([`590456e`](https://github.com/bitcoin/bitcoin/commit/590456e3f1043ba0680e0afec9fd7653db1098bb)); this port did not.

Because **no functional tests run in this repo's CI** (macOS and Win64 jobs are `if: false`, see #203), neither failure will surface on its own — someone will hit it running the suite by hand.

**Fix:** add `-mempoolfullrbf=0` to `mempool_accept.py`'s node args; in `p2p_permissions.py` make the tx fail independently of RBF signaling, as upstream did (append a zero-value dust output, expect `dust`).

### [medium] Nothing tests full-RBF as the *unset* default

`test/functional/feature_rbf.py:36` — this PR adds `-mempoolfullrbf=0` to **both** nodes, and `test_fullrbf()` later restarts with `-mempoolfullrbf=1`. Every mode the test exercises is now selected explicitly, so **the suite passes identically whether `DEFAULT_MEMPOOL_FULL_RBF` is `false` or `true`** — the one behaviour this PR exists to change is untested.

**Fix:** in `test_fullrbf()`, restart with `extra_args=[]` and assert `getmempoolinfo()["fullrbf"]` is true. Better still, add a unit assertion in `mempool_args_tests.cpp` that an unset option yields `full_rbf=true` and `-mempoolfullrbf=0` overrides it — that one runs in CI.

### [low] Packaged docs still state the old default

The policy page in this diff is correct, but three shipped files still say `(default: 0)`:

- `doc/man/btqd.1:677`
- `doc/man/btq-qt.1:677`
- `share/examples/btq.conf:569`

Runtime help already prints `1` (it uses `strprintf` with the constant), so packaged docs now contradict the binary. Regenerate the manpages and the example config, and add a release note recording the 0 → 1 transition — operators with zero-confirmation assumptions need to see it.

### Verified correct

Policy-only: the value flows `MemPoolOptions::full_rbf` → `CTxMemPool::m_full_rbf` → the mempool admission path in `validation.cpp`, and touches no block-validity rule. During mixed rollout, opt-in and full-RBF nodes will have diverging mempools and topology-dependent relay, but both accept the resulting blocks — so no coordinated activation is needed.

BarneyChambers:
Full RBF: a stuck unconfirmed tx can be replaced by a higher-fee version even if the original did not set BIP125 (sequence below SEQUENCE_FINAL). Default is on, matching Core 28. `-mempoolfullrbf=0` still restores opt-in.

Two functional tests still expected `txn-mempool-conflict` for an opt-out replacement that pays more. Under full-RBF that replacement is accepted or fails later on fee. `mempool_accept.py` now starts with `-mempoolfullrbf=0` so the signaling case still means something. `p2p_permissions.py` appends a dust output and expects `dust`, which does not depend on RBF. `test_fullrbf` restarts with no extra args so it actually exercises the unset default. Manpages and the example config still said default 0.

augchan42:
## Follow-up review — the replacement test assertion is still wrong

The RBF default and the doc updates look right. The new test line does not.

`test/functional/p2p_permissions.py:116` appends a zero-value bare `OP_TRUE` output and expects the reject reason `dust`. It will get `scriptpubkey` instead. In `IsStandardTx`, the output loop calls `::IsStandard(txout.scriptPubKey, ...)` first and returns on a non-standard script; the `IsDust` check sits in the `else` branch below it. A bare `OP_TRUE` solves to `NONSTANDARD`, so it never reaches the dust check.

Use a zero-value output with a standard script — a P2WPKH scriptPubKey will do.


---

augchan42:

The dust fix is right. `key_to_p2wpkh_script(bytes(33))` gives a standard P2WPKH script, so `IsStandardTx` (`src/policy/policy.cpp:137-148`) gets past the script check and rejects the zero-value output as `dust`. The second send still hits `m_recent_rejects`.

The earlier findings are all addressed:

- `mempool_accept.py` runs with `-mempoolfullrbf=0`.
- `test_fullrbf` restarts node 0 with no extra args and checks that `fullrbf` is true, so the unset default is tested. It is the last test in the file, so dropping node 0's other args there affects nothing.
- `mempool_args_tests` checks the new default.
- The manpages and `btq.conf` say default 1.

Tested:

- master + this PR: `p2p_permissions.py` passes. `feature_rbf.py` and `mempool_accept.py` fail with the same subsidy errors they fail with on master. Both are in `known_failures.txt`.
- master + this PR + #212–#215, which fix those subsidy errors: `feature_rbf.py`, `mempool_accept.py`, `p2p_permissions.py` and `mempool_args_tests` pass. The full functional suite gives 256 passed and 39 failed, all 39 on the known-failures list with none new. That matches #212–#215 without this PR, so the new default breaks no other test.

This PR merges cleanly with master and with #212–#215. Both it and #212 edit `feature_rbf.py` and `mempool_accept.py`, in different places.

Not for this PR: the default change matters to anyone who accepts zero-confirmation payments, so list it in the next release notes.

---

augchan42 (approved):
Approving. Details in the follow-up review above.

augchan42:
Approved and ready to merge whenever you want to land it. Nothing further from my side. #184 and #185 also change `src/kernel/mempool_options.h`, so pick the merge order that suits you.

PR #189, opened 2026-09-01
validation: stop using peer-adjusted time for consensus paths

BarneyChambers:
## Summary
- Bitcoin Core 27.0 #28956: header and template time no longer follow the peer median.
- `ContextualCheckBlockHeader` already takes a `now` callback. It is now `NodeClock::now` instead of `GetAdjustedTime`.
- `AddTimeData` and the 70-minute clock warning stay. Operators still need an honest system clock. That matters more here: 60s blocks and a 15-minute future window.

## Test plan
- [ ] `src/test/test_btq --run_test=timedata_tests`
- [ ] `src/test/test_btq --run_test=miner_tests`
- [ ] Mine a regtest block and check `nTime` is wall-clock, not peer-skewed
- [ ] Confirm a 70-minute peer offset still logs the warning

### DIFF
--- a/share/examples/btq.conf
+++ b/share/examples/btq.conf
@@ -242,9 +242,9 @@
 # (default: 1000)
 #maxsendbuffer=<n>
 
-# Maximum allowed median peer time offset adjustment. Local perspective of
-# time may be influenced by outbound peers forward or backward by
-# this amount (default: 4200 seconds).
+# Maximum allowed median peer time offset adjustment. Used only for
+# clock-offset warnings; block and template time use the system clock
+# (default: 900 seconds).
 #maxtimeadjustment=1
 
 # Tries to keep outbound traffic under the given target per 24h. Limit
--- a/share/examples/btq.conf
+++ b/share/examples/btq.conf
@@ -244,7 +244,7 @@
 
 # Maximum allowed median peer time offset adjustment. Used only for
 # clock-offset warnings; block and template time use the system clock
-# (default: 900 seconds).
+# (default: 4200 seconds).
 #maxtimeadjustment=1
 
 # Tries to keep outbound traffic under the given target per 24h. Limit
--- a/src/headerssync.cpp
+++ b/src/headerssync.cpp
@@ -5,7 +5,7 @@
 #include <headerssync.h>
 #include <logging.h>
 #include <pow.h>
-#include <timedata.h>
+#include <util/time.h>
 #include <util/check.h>
 #include <util/vector.h>
 
@@ -41,7 +41,7 @@ HeadersSyncState::HeadersSyncState(NodeId id, const Consensus::Params& consensus
     // exceeds this bound, because it's not possible for a consensus-valid
     // chain to be longer than this (at the current time -- in the future we
     // could try again, if necessary, to sync a longer chain).
-    m_max_commitments = 6*(Ticks<std::chrono::seconds>(GetAdjustedTime() - NodeSeconds{std::chrono::seconds{chain_start->GetMedianTimePast()}}) + MAX_FUTURE_BLOCK_TIME) / HEADER_COMMITMENT_PERIOD;
+    m_max_commitments = 6*(Ticks<std::chrono::seconds>(NodeClock::now() - NodeSeconds{std::chrono::seconds{chain_start->GetMedianTimePast()}}) + MAX_FUTURE_BLOCK_TIME) / HEADER_COMMITMENT_PERIOD;
 
     LogPrint(BCLog::NET, "Initial headers sync started with peer=%d: height=%i, max_commitments=%i, min_work=%s\n", m_id, m_current_height, m_max_commitments, m_minimum_required_work.ToString());
 }
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -493,7 +493,7 @@ void SetupServerArgs(ArgsManager& argsman)
     argsman.AddArg("-maxconnections=<n>", strprintf("Maintain at most <n> connections to peers (default: %u). This limit does not apply to connections manually added via -addnode or the addnode RPC, which have a separate limit of %u.", DEFAULT_MAX_PEER_CONNECTIONS, MAX_ADDNODE_CONNECTIONS), ArgsManager::ALLOW_ANY, OptionsCategory::CONNECTION);
     argsman.AddArg("-maxreceivebuffer=<n>", strprintf("Maximum per-connection receive buffer, <n>*1000 bytes (default: %u)", DEFAULT_MAXRECEIVEBUFFER), ArgsManager::ALLOW_ANY, OptionsCategory::CONNECTION);
     argsman.AddArg("-maxsendbuffer=<n>", strprintf("Maximum per-connection memory usage for the send buffer, <n>*1000 bytes (default: %u)", DEFAULT_MAXSENDBUFFER), ArgsManager::ALLOW_ANY, OptionsCategory::CONNECTION);
-    argsman.AddArg("-maxtimeadjustment", strprintf("Maximum allowed median peer time offset adjustment. Local perspective of time may be influenced by outbound peers forward or backward by this amount (default: %u seconds).", DEFAULT_MAX_TIME_ADJUSTMENT), ArgsManager::ALLOW_ANY, OptionsCategory::CONNECTION);
+    argsman.AddArg("-maxtimeadjustment", strprintf("Maximum allowed median peer time offset adjustment. Used only for clock-offset warnings; block and template time use the system clock (default: %u seconds).", DEFAULT_MAX_TIME_ADJUSTMENT), ArgsManager::ALLOW_ANY, OptionsCategory::CONNECTION);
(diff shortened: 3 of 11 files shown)


BarneyChambers:
Tracking: #197

augchan42:
## Independent review — 4 findings (1 high, 2 medium, 1 low)

Cross-model review (Codex, context-blind) against `.claude/code-review-lanes.md`. Every finding verified against the code before posting. The change itself is a faithful backport of upstream's removal of adjusted time; the issue is BTQ-specific rollout, not the port.

### [high] Ungated clock switch can split patched and unpatched nodes mid-testnet

Header acceptance moves from `H <= system_time + peer_offset + MAX_FUTURE_BLOCK_TIME` to `H <= system_time + MAX_FUTURE_BLOCK_TIME`, with no activation height.

This is riskier in BTQ than upstream because of one constant BTQ changed and one it didn't:

| Constant | Bitcoin Core | BTQ | Where |
|---|---|---|---|
| `MAX_FUTURE_BLOCK_TIME` | 7200 s | **900 s** | `src/chain.h:24` |
| `DEFAULT_MAX_TIME_ADJUSTMENT` | 4200 s | **4200 s** (inherited) | `src/timedata.h:16` |

Upstream has adjustment (4200) **below** allowance (7200), so dropping adjusted time narrows an already-larger window. **BTQ has the relationship inverted — the peer adjustment window is 4.7x the future-block allowance.**

Concretely: an old node carrying a `+4200` offset accepts a header at `T+5100` that a patched node rejects until `T+3300` — a ~55 minute disagreement, sustainable by a miner that maintains the offset.

`BLOCK_TIME_FUTURE` is not cached, so this is recoverable rather than a permanent fork. It is still a live-testnet hazard.

**Suggested:** coordinate the testnet upgrade, or reduce `DEFAULT_MAX_TIME_ADJUSTMENT` to at most 900 s in a preceding commit so the two windows are consistent before the switch.

### [medium] No test distinguishes adjusted-time from system-time validation

`src/test/util/setup_common.cpp:183` — the test changes only swap fixture callbacks. No test sets a nonzero peer offset, so the existing future-time tests pass identically before and after this patch. Nothing would catch a missed call site or the 900/901-second boundary. Note no functional tests run in CI (see #203), so a C++ unit test is the higher-value addition here.

### [low] `-maxtimeadjustment` now advertises behaviour it no longer has

After this change `GetAdjustedTime()` has **zero** callers outside `src/timedata.*` (verified). The option still bounds offset reporting and the clock warning, but `src/init.cpp:496` still reads:

> Local perspective of time may be influenced by outbound peers forward or backward by this amount

That is now false. `ChainstateManager::Options::adjusted_time_callback` and the `chain.h` comments likewise still describe adjusted time, which invites a future caller to reintroduce peer time into validation. Consider renaming the callback to `current_time_callback` and rewording the help to say it affects only offset reporting and warnings.

### Filed separately, not for this PR

The `m_max_commitments` computation at `src/headerssync.cpp:44` can wrap to a huge `uint64_t` when the numerator goes negative. **Pre-existing** — the same expression is on master with `GetAdjustedTime` and in upstream Core, so it is not introduced here. Filed as its own issue; please don't fix it inside this PR.

BarneyChambers:
Header and template time no longer follow the peer median. They use the system clock. `-maxtimeadjustment` is still there for the 70-minute clock warning, it does not move consensus.

BTQ already cut `MAX_FUTURE_BLOCK_TIME` to 15 minutes. The peer offset cap stayed at 70 minutes, Bitcoin's number. Relationship is inverted versus Core, where the adjustment (4200s) is smaller than the future window (7200s). An old node carrying a +4200s offset would accept a header this node rejects until wall clock catches up, about 55 minutes of disagreement on a 60s chain.

`DEFAULT_MAX_TIME_ADJUSTMENT` is 900s now, same as the future-block window. Help text says it only bounds offset reporting. Unit test asserts the two stay ordered.

augchan42:
## Follow-up review — revised; the clock-source switch is fine, the constant is not

**Correction first.** I previously said this needs an explicit testnet transition and old-miner decision. That was wrong — the decision is already made. #197 settles it: "This is not a hard fork... Temporary skew, not a flag day." #144 also resets testnet from a new genesis after the audit, which closes the mixed-version window. Please disregard that ask.

The clock-source switch is right and matches #197: `NodeClock::now` into the header check, the miner template, `UpdateTime`, and the IBD paths, with `AddTimeData` left in place. No concerns there.

### High — restore `DEFAULT_MAX_TIME_ADJUSTMENT` to 4200

`src/timedata.h` drops it from 4200 to 900 with this rationale:

> A larger peer offset would let an unpatched node accept headers a patched node rejects.

That does not work. An unpatched node compiles in its own 4200 and reads its own `-maxtimeadjustment`. Lowering the constant in the new binary has no effect on old nodes, which is exactly the case the comment is trying to cover.

What it does affect is the operator warning. After this branch, no consensus or mining path reads `nTimeOffset` — `src/init.cpp:1450` passes `NodeClock::now` instead. The constant's only remaining consumer is `src/timedata.cpp:81`, where it decides when we tell an operator their clock is wrong. Lowering it to 900 narrows that warning from 70 minutes to 15.

#197 rules that out directly — "`AddTimeData` and the 70-minute warning should stay" — and lists as a check: "A 70-minute peer offset still logs the warning."

Please restore 4200 and drop the `timedata.h` comment. Keep the `btq.conf` and `init.cpp` help text changes: "used only for clock-offset warnings" is now accurate.

### Medium — the test does not exercise the change

`max_time_adjustment_fits_future_block_window` asserts one constant against another. It passes whatever the wiring does. Something that exercises the change would be better: inject a peer offset, then confirm a mined template's `nTime` and `ContextualCheckBlockHeader`'s future-time boundary both track the system clock and ignore the offset.


---
_Revised 2026-09-09 after several review rounds across Opus and Codex, context-blind against the current head. The correction above supersedes what this comment said earlier._

augchan42:
**Review finding (Medium): the clock warning threshold is looser than the future-block window**

- `DEFAULT_MAX_TIME_ADJUSTMENT` stays at 70 min (`src/timedata.h:16`); commit dede7d635 reverted the 15 min change and deleted the test that tied the two values together.
- `MAX_FUTURE_BLOCK_TIME` is 15 min (`src/chain.h:24`), and consensus paths now compare against the unadjusted system clock.

A node whose clock is 15 to 70 min off is therefore broken with no warning. `AddTimeData` treats that offset as within bounds, so `SetMiscWarning` never fires. There are two ways this shows up:
- **Slow clock:** every honest header is rejected as `time-too-new` and the node stalls at its tip.
- **Fast clock (miner):** `UpdateTime`/`CreateNewBlock` stamp blocks 15+ min ahead, and correctly-clocked peers reject them.

Before this PR the peer offset corrected both cases. This isn't a consensus split: `BLOCK_TIME_FUTURE` isn't cached as failed and isn't punished, so the node recovers once its clock is fixed. It is a liveness and wasted-hashpower issue with no operator signal.

Upstream made the same change with a 2h window and lowered its warning to 10 min, well inside it. Suggested fix: warn at or below `MAX_FUTURE_BLOCK_TIME` (e.g. 10 min) and restore a `static_assert` or test tying the two values together.

Nit: comments at `chain.h` ("network-adjusted time"), `net_processing.cpp:~600` and `validation.cpp:~2208` ("2 hours") are stale.

_Checked by Opus._

BarneyChambers:
DEFAULT_MAX_TIME_ADJUSTMENT is 10 minutes, inside MAX_FUTURE_BLOCK_TIME.

A node 15-70 minutes off used to stall or mine rejected blocks with no SetMiscWarning, because the warning still used the 70-minute peer-offset cap. It now warns at 10 minutes. timedata_tests checks the warning cap is <= the future-block window.

block_time_ignores_peer_offset now uses a 10-minute peer offset so AddTimeData actually applies it. Comments that still said 2 hours / network-adjusted time are updated. pruneblockchain still subtracts TIMESTAMP_WINDOW (15 minutes).

BarneyChambers:
Pushed a follow-up so packaged help matches the code:

- `btq.conf` and manpages now say default 600 seconds, not 4200.
- Manpages match `init.cpp`: warnings only; block and template time use the system clock.

augchan42:
## Follow-up review: clean

**Fixed:**
- **Warning threshold:** `DEFAULT_MAX_TIME_ADJUSTMENT` is 10 minutes. Any median offset above that enters the warning branch. `addtimedata` tests 601 s and checks the warning fires. `timedata_tests.cpp:106-109` ties the cap to `MAX_FUTURE_BLOCK_TIME`.
- **Offset no longer used:** `GetAdjustedTime` has no production callers, and `GetTimeOffset` is only reported by `getnetworkinfo`.
- **New test:** `block_time_ignores_peer_offset` distinguishes adjusted time from system time: +900 is accepted and +901 is rejected with `time-too-new`, whatever the peer offset.
- **Docs:** manpages, `btq.conf` and the "2 hours" comments are updated.

**Chain-split concern (Tier 1):** there is still no activation height. Lowering the default doesn't change unpatched nodes, which keep 4200. That's covered by the #197 decision and the testnet reset, and `BLOCK_TIME_FUTURE` isn't cached as a failure.

**Notes, all predating this PR:**
- One outbound peer reporting a nonzero offset under 5 minutes suppresses the warning (`fMatch`, `timedata.cpp:91`), the same as upstream 26.
- The #197 "Check" section still says 70 minutes.
- Two other issues come from the 15-minute window that was already on master: `contrib/signet/miner` still allows blocks up to 6900 s in the future, and three functional tests assume a 2 h `TIMESTAMP_WINDOW`. Filed as #223; not for this PR.

_Checked by Codex and Opus._

PR #192, opened 2026-09-01
wallet: add CoinGrinder for high-feerate Dilithium sends

BarneyChambers:
## Summary
- Port CoinGrinder from Bitcoin Core 27.0 #27877.
- It runs when the effective feerate is more than 2× the long-term feerate (Core uses 3×). Dilithium inputs are huge; grind earlier.
- BnB, knapsack, and SRD still run. The least-waste result wins.

## Test plan
- [ ] `src/test/test_btq --run_test=coinselector_tests`
- [ ] Send a high-feerate Dilithium tx from a wallet with many small UTXOs and confirm input count is not worse than before

### DIFF
--- a/src/wallet/coinselection.cpp
+++ b/src/wallet/coinselection.cpp
@@ -13,6 +13,8 @@
 #include <util/check.h>
 #include <util/moneystr.h>
 
+#include <algorithm>
+#include <limits>
 #include <numeric>
 #include <optional>
 #include <queue>
@@ -32,6 +34,17 @@ struct {
         return a.GetSelectionAmount() > b.GetSelectionAmount();
     }
 } descending;
+
+// Sort by descending effective value; lower weight wins a tie.
+struct {
+    bool operator()(const OutputGroup& a, const OutputGroup& b) const
+    {
+        if (a.GetSelectionAmount() == b.GetSelectionAmount()) {
+            return a.m_weight < b.m_weight;
+        }
+        return a.GetSelectionAmount() > b.GetSelectionAmount();
+    }
+} descending_effval_weight;
 
 /*
  * This is the Branch and Bound Coin Selection algorithm designed by Murch. It searches for an input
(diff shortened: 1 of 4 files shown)


BarneyChambers:
Tracking: #200

augchan42:
## Independent review — 2 findings, both about bounds and coverage

The port itself looks faithful to upstream. Two things worth addressing, and one claim I checked and **rejected** before posting — noted at the end so it isn't raised again later.

### [medium] The iteration cap doesn't bound the clone-skipping loop

`src/wallet/coinselection.cpp:387`. `TOTAL_TRIES = 100,000` bounds *evaluated selections* through `curr_try`, but the inner loop that skips equal-effective-value UTXOs is not charged against it:

```cpp
while (utxo_pool[next_utxo - 1].GetSelectionAmount() == utxo_pool[next_utxo].GetSelectionAmount()) {
    if (next_utxo >= utxo_pool.size() - 1) { should_shift = true; break; }
    ++next_utxo;   // not counted against TOTAL_TRIES
}
```

Worst case is therefore `O(TOTAL_TRIES x N)` — each of up to ~100,000 iterations can rescan a long suffix of clones. Transaction creation holds wallet state throughout, so this is a wallet-hang shape rather than just a slow path.

**To be fair about attribution:** this matches upstream Bitcoin Core verbatim, so it isn't a BTQ implementation mistake, and I'd not treat it as a merge blocker on those grounds. It is more reachable here, though — the PR's own motivation is Dilithium wallets, where large runs of equal-value multi-kilobyte UTXOs are exactly the input that triggers it.

**Fix if you want it bounded:** precompute in `O(N)` the next index with a differing effective value and jump to it in `O(1)`, or charge skipped clones against the total-work budget. Otherwise, worth a comment recording that the bound is on evaluated selections, not total work.

### [medium] The test runs the weight-minimising algorithm with no weights

`src/wallet/test/coinselector_tests.cpp:190`. The `add_coin` helper used here sets `input_bytes = -1`, so every `OutputGroup::m_weight` is **zero**. CoinGrinder exists to minimise weight, and the one new test gives it nothing to minimise.

The assertions are the algorithm tag and `selected_value >= 3*CENT`. An implementation that ignores the 3.1-cent change target, picks a non-minimal input set, mishandles `max_weight`, or never reports iteration exhaustion passes all of them.

Upstream's CoinGrinder coverage — insufficient funds, max weight, mixed weights, lighter combinations, clones, tiny UTXOs, iteration exhaustion, plus fuzz — wasn't ported alongside the algorithm.

**Fix:** port upstream's unit and fuzz cases, then add a BTQ-specific one using real `CalculateMaximumSignedInputSize` results for P2MR/Dilithium inputs, asserting the selected set, its weight, the change target, and the 100,000-try exhaustion path. These run under `make check`, so they'd actually gate CI.

### Checked and rejected

The review also claimed, at [high], that `BOOST_CHECK_EQUAL(result->GetAlgo(), SelectionAlgorithm::CG)` cannot compile because `SelectionAlgorithm` is a scoped enum with no `operator<<`. **That is false**, and I'm recording it here so it doesn't resurface: CI run [33514087871](https://github.com/btq-ag/btq-core/actions/runs/33514087871) on exactly this head (`da127597`) shows

```
CXX      wallet/test/test_test_btq-coinselector_tests.o
...
Running tests: coinselector_tests from wallet/test/coinselector_tests.cpp
```

and the job passed. `make check` is a step in that job, so the file compiles and the test runs. No change needed.

BarneyChambers:
CoinGrinder is the coin-selection pass that tries to spend as little witness weight as possible when the feerate is high. Dilithium inputs are several kilobytes, so picking a heavy UTXO set hurts more than it does on Bitcoin.

The new test set `input_bytes = -1` on every coin, which makes `OutputGroup::m_weight` zero. The algorithm had nothing to minimise. Assertions were "it returned CG" and "value >= 3 CENT". A version that ignored weight, max_weight, or the change target would still pass.

The test now has two 148-byte inputs vs one 4000-byte one, target 3 CENT, and checks we take the light pair.

The inner loop that skips equal-value clones is still not charged against the 100k try cap, same as Core. I left a comment rather than changing the port. Worst case is still `O(tries * N)` while the wallet is locked.

augchan42:
## Follow-up review — test fixed; bound acknowledged

The revised test is right: real input weights (148 vs 4000 bytes), and it asserts the two lighter inputs are selected.

The equal-value clone skip is not charged to `TOTAL_TRIES`, so a long run of equal-value UTXOs can still do O(tries x N) work under the wallet lock. You have already written that down in the code comment, and it is inherited from upstream CoinGrinder, so I am not asking you to change it here. Noting it so the next person reading that loop finds it flagged.


---

augchan42:

The head (2cb0d7839) is the one I reviewed on Sep 8. Nothing has changed since.

**Compared with upstream.** Ignoring whitespace, `CoinGrinder` is identical to Bitcoin Core v27.0 apart from the comment added to the clone-skipping loop. The tie-break comparator, the enum values (`CG = 3`, `MANUAL = 4`), the completed and evaluated counters, and the call in `ChooseSelectionResult` match too. The enum value is never stored; only its name reaches the tracepoint. Upstream's later changes to the function (v28.0 and v29.0) rename `max_weight` and remove one dead assignment, so the port isn't missing a fix.

**The one deliberate difference** is the trigger: 2× the long-term feerate instead of 3×. With the default `-consolidatefeerate` of 10 sat/vB, CoinGrinder runs above 20 sat/vB instead of 30. The code comment and commit message both say so. The least-waste result still wins, so between 20 and 30 sat/vB this adds one more candidate. It costs up to 100,000 tries on those sends and can't make the chosen result worse by the waste measure.

**Dilithium sizing.** CoinGrinder minimises `OutputGroup::m_weight`, which is `input_bytes × 16`. P2MR coins get a real `input_bytes` from `DummySignInputWeight` in `spend.cpp`: 41 non-witness bytes × 16, plus the witness. That's about 276 vB for a single-leaf Dilithium input, so CoinGrinder sees P2MR coins at their real weight, not zero.

Two small things, neither blocking:

- In the clone-skipping loop, the added comment and `++next_utxo;` are indented at the level of the `while`, so the increment reads as if it were outside the loop. The code is correct. Indenting it as upstream does also keeps later diffs against upstream clean.
- The test comment calls a 4000-byte input "Dilithium-sized". `input_bytes` is in vbytes, and a single-leaf P2MR input is about 276 vB, so 4000 is about 14 times a real one. The test is still valid; only the label is off.

Tested:

- Built on the #212–#215 stack with this PR and #184. `coinselector_tests` passes, including `coingrinder_min_weight_test`, and so do the wallet, spend and IsMine unit suites.
- Full functional suite with `check_known_failures.py`: 304 tests, 257 passed, 38 failed (all 38 on the known-failures list, none new), 9 skipped.
- A regtest script built one wallet with 35 P2MR and 23 P2WPKH coins, then funded the same three payments 15 times each at 5, 15, 25, 50 and 100 sat/vB, with and without this PR. Nothing was broadcast, so every attempt saw the same coins. Above 20 sat/vB, this PR picked the lightest set every time:

  | Payment | Without this PR (25–100 sat/vB) | With this PR |
  |---|---|---|
  | 0.35 | 1 or 2 inputs, 149–152 vB on average | 1 input, 120 vB |
  | 2.5 | usually 3 inputs, 213–216 vB | 2 inputs, 168 vB |
  | 6 | usually 9 inputs, 492–502 vB | 4 inputs, 263 vB |

  At 5 and 15 sat/vB, CoinGrinder doesn't run and the two builds gave the same results. The same script confirms P2MR inputs are sized: at 5 sat/vB the wallet consolidated 26 P2MR coins in 7,226 vB, about 276 vB each.
- In the functional tests, the wallet's fee estimates come out just above 20 sat/vB, for example 2,401 sat for a 120 vB send. So with this PR, CoinGrinder runs for many test sends; at 3× it never would. The suite has no new failures. One side effect: `wallet_avoidreuse.py --descriptors`, on the known-failures list because it hits the 25-transaction mempool chain limit, passed 7 of 14 runs with this PR and 0 of 13 without it. It should stay on the list until it passes every time.

The clone walk still isn't charged to `TOTAL_TRIES`, as noted last time. That matches upstream, and the code comment records it.

---

PR #193, opened 2026-09-01
policy: retune fee estimator for 60-second blocks

BarneyChambers:
## Summary
- Fee estimator horizons still assumed 10-minute blocks. Short/med/long now track ~2 hours, ~8 hours, and ~1 week of 60s blocks, with matching decay half-lives.
- `estimatesmartfee` / `-txconfirmtarget` docs say conf_target=6 is about six minutes. Qt send targets are 2, 6, 12, 30, 60, 180, 720, 1440.
- `fee_estimates.dat` format 2. Old 10-minute files are discarded on read.

## Test plan
- [ ] `src/test/test_btq --run_test=policyestimator_tests`
- [ ] `test/functional/rpc_estimatefee.py`
- [ ] After ~1000 regtest blocks, `estimatesmartfee 6` returns a feerate (not only the fallback)
- [ ] Restart loads a format-2 `fee_estimates.dat`; a pre-change file is ignored


BarneyChambers:
Tracking: #201

augchan42:
## Independent review — the decay retune is right; two per-block rates were left behind

**The decay work is sound.** Half-lives (`H = ln(0.5)/ln(decay)`) move 17.89→180.16, 144.06→1443.71, and 1004.21→10045.26 blocks — factors of 10.07, 10.02, 10.00. In wall-clock minutes each horizon is preserved to within 0.7%. The scale/period products, `OLDEST_ESTIMATE_HISTORY`, the fee-file max confirmations and the RPC bounds all scale cleanly by 10.

The findings are the constants that *didn't* move with them.

### [medium] Two sufficiency thresholds are per-block and were not scaled

`src/policy/fees.h:180,182` — both unchanged by this PR:

```cpp
static constexpr double SUFFICIENT_FEETXS = 0.1;
static constexpr double SUFFICIENT_TXS_SHORT = 0.5;
```

These are transactions **per block**, and the acceptance threshold is `sufficientTxVal / (1 - decay)`. Retuning the decays by 10x therefore multiplies the required transaction count by 10:

| Horizon | Before | After |
|---|---|---|
| Short | `0.5 / (1-0.962)` = 13.16 | `0.5 / (1-0.99616)` = **130.21** |
| Medium | `0.1 / 0.0048` = 20.83 | `0.1 / 0.00048` = **208.33** |
| Long | `0.1 / 0.00069` = 144.93 | `0.1 / 0.000069` = **1449.28** |

The same wall-clock history now needs ten times as many transactions to produce an estimate. Concretely, one transaction every 50 minutes is 0.2 tx/block upstream and clears the 0.1 threshold; on BTQ it is 0.02 tx/block and fails. On a young/low-volume chain this is exactly when you most want an estimate.

**Fix:** scale to wall-clock equivalents — `SUFFICIENT_FEETXS = 0.01`, `SUFFICIENT_TXS_SHORT = 0.05` — and add low-volume regression coverage.

### [medium] The short horizon scales a different dimension than its siblings

`src/policy/fees.h:154-161`:

| Horizon | Periods | Scale | Bucket width | Product |
|---|---|---|---|---|
| Short | 12 → **120** | 1 → **1** | 10 min → **1 min** | 120 ✓ |
| Medium | 24 → 24 | 2 → **20** | 20 min → 20 min | 480 ✓ |
| Long | 42 → 42 | 24 → **240** | 240 min → 240 min | 10,080 ✓ |

All three products grow 10x, but short got there by multiplying *periods* while the others multiplied *scale*. So medium and long keep their wall-clock bucket width and short's drops from 10 minutes to 1 — different statistical resolution and 10x the confirmation/failure arrays, rather than the same estimator retuned.

**Fix:** `SHORT_BLOCK_PERIODS = 12`, `SHORT_SCALE = 10`. If 1-minute buckets are deliberate, say so and add short-horizon tests for it.

### [medium] Wallet and GUI fee targets are not wall-clock equivalent

`DEFAULT_TX_CONFIRM_TARGET` is still **6** (`src/wallet/wallet.h:131`) — one hour upstream, **six minutes** here. The GUI list in `src/qt/sendcoinsdialog.cpp:45` went `{2,4,6,12,24,48,144,504,1008}` → `{2,6,12,30,60,180,720,1440}`: positional factors of 1, 1.5, 2, 2.5, 2.5, 3.75, 5, 2.857, with the week-long target dropped. Preserving the old wall-clock choices would be `{20,40,60,120,240,480,1440,5040,10080}`.

This changes the fee policy the wallet actually applies, not just the estimator's history. Worth noting `src/rpc/fees.cpp:35` already says *"six minutes, not an hour"* — the help text acknowledges the shift that the default was not adjusted for. If a 6-minute default is intended, that's a defensible call, but it should be a stated decision rather than a leftover.

### [low] The tests were weakened rather than retargeted

`src/test/policyestimator_tests.cpp:128` replaces exact scale-aligned fee expectations with `estimate > 0`, while still adding four transactions per block — ten times the previous wall-clock rate. The test now passes under an incorrect scale-to-target mapping. Functional changes only swap the rejected sentinel 1,009 → 10,081; nothing asserts 10,080 is *accepted*, and nothing covers the fee-file format marker change (149,900 → 2) surviving a restart.

### [low] RPC help claims BIP141 virtual-size semantics

`src/rpc/fees.cpp:35` and `:109` say fee rates use virtual size "as defined in BIP 141". BIP141 divides weight by 4; BTQ's `GetVirtualTransactionSize` divides by `WITNESS_SCALE_FACTOR = 16`. That's a 4x difference, and it matters most for the witness-heavy Dilithium transactions this PR is tuning for. Describe it as BTQ virtual bytes at WSF 16 instead.

### No findings

Consensus divergence, resource limits, security: nothing. No changed value affects mempool admission, relay, or mining selection — this is estimation, persistence and wallet fee choice only.

BarneyChambers:
The fee estimator watches how long unconfirmed txs sit, then answers what feerate gets you into a block in N blocks. Bitcoin sized those horizons for 10-minute blocks. We already retuned decay and bucket counts for 60-second blocks.

Two thresholds did not move: how many txs per block count as "enough data" to emit an estimate. After the decay change those required counts went up 10x, so a quiet chain would sit on the fallback fee for much longer. They are now 0.01 and 0.05, the wall-clock equivalent of Core's 0.1 / 0.5.

The short horizon had gotten to 10x by multiplying periods (120 x scale 1), which made 1-minute buckets while medium and long kept their old widths. It is now 12 periods x scale 10, same shape as the others.

Wallet default confirm target was still 6, which is six minutes here and was one hour on Bitcoin. It is 60. Qt's target list is the old wall-clock set. Fee RPC help said BIP141 virtual size (weight / 4). Ours is weight / 16.

augchan42:
## Follow-up review — two stale strings left

The estimator corrections match 60-second block timing. Two documented defaults were left behind.

`src/wallet/wallet.h:132` now sets `DEFAULT_TX_CONFIRM_TARGET` to 60, but both places that document it still say 6:

- `doc/man/btqd.1:489`
- `share/examples/btq.conf:408`

Please update both.


---

augchan42:
## Follow-up review — substantive tuning fixes verified

The short horizon, sufficiency thresholds, wallet/Qt targets, and BTQ weight/16 RPC wording now match the intended 60-second timing. Those earlier findings are resolved.

One documentation mismatch remains: runtime `DEFAULT_TX_CONFIRM_TARGET` is now **60**, while `doc/man/btqd.1` and `share/examples/btq.conf` still advertise **6**. Please update/regenerate both before merge.

Low-volume estimator behavior, exact target mapping, the accepted upper boundary, and fee-file restart/migration remain coverage gaps rather than confirmed runtime defects.

augchan42:
**Review findings**

**1. Medium: unmined removals under 10 blocks old are no longer counted as failures**

`SHORT_SCALE` goes from 1 to 10 (`src/policy/fees.h:157`), but `TxConfirmStats::removeTx` still only records a failure when `blocksAgo >= scale` (`src/policy/fees.cpp:496`). The smallest scale is now 10, so a tx that sits unmined for 1-9 blocks and is then replaced or evicted leaves no failure in any horizon. Its fee bucket looks like 100% success.

`sanity_check_rbf_estimates` in `feature_fee_estimation.py` sets up exactly this case. 225 of 250 txs at 1 sat/vB are replaced one block later. The short and medium horizons then see 25 confirmed / 0 failed at 1 sat/vB, so `estimatesmartfee 2 economical` returns about 1 sat/vB instead of about 10. The test's conservative call only passes because the long horizon lacks enough data points. On mainnet, anything RBF'd or evicted within about 10 minutes drags economical estimates down.

Upstream only ignores removals before a single block has passed. One option is to track failures in block units: compare `blocksAgo >= 1` and bucket by `blocksAgo / scale`.

**2. Low: saved Qt confirmation targets are silently remapped**

`nConfTarget` is stored in QSettings as a raw block count from the old list `{2..1008}`. `src/qt/sendcoinsdialog.cpp` now maps it through the new `confTargets{20,...,10080}` without migrating it. A saved 1008 ("1 week") becomes 1440 (1 day), 504 also becomes 1440, 144 becomes 240, and 2/4/6/12 become 20. Users end up paying higher fees without noticing.

**Unverified / minor**
- `(nBlockHeight - confct) % bins` in `EstimateMedianVal` wraps as an unsigned value when height < confct. With 10080 bins this now affects the first week of chain and all of regtest (it used to be the first 1008 blocks).
- The 10080-row long horizon makes `EstimateMedianVal` walk up to about 1.9M cells for the default wallet target. That's a performance cost, not a correctness bug.

The file format handling is fine: old `fee_estimates.dat` files are rejected cleanly by the format marker and are never misparsed.

_Checked by Opus._

augchan42:
#221 covers `ROLLING_FEE_HALFLIFE`, which this PR doesn't touch. It's 12 hours in `src/txmempool.h:330`, upstream's value for 10-minute blocks. At 60 seconds that's 720 blocks instead of 72.

It sets the mempool's eviction fee floor, not the estimator, so it's a separate code path. The cause is the same as the half-lives here. Worth deciding whether it belongs in this PR or stays separate.

BarneyChambers:
Unmined removals after one block now count as estimator failures.

SHORT_SCALE is 10, so the old blocksAgo >= scale skip left 1-9 block RBF/eviction looking like 100% success. removeTx now records a failure when blocksAgo >= 1 and buckets by max(1, blocksAgo/scale). Same-block removals are still ignored.

Qt nConfTarget values from the old {2..1008} list are multiplied by 10 on load so a saved 1008 stays "1 week" instead of mapping to 1440. Already-new values are left alone.

ROLLING_FEE_HALFLIFE is still 12 hours. That is the mempool eviction floor, not the estimator, so it stays on #221.

BarneyChambers:
Merged current master.

CI's new failure was `wallet_basic.py --legacy-wallet`: verbose send returned "Half Target 60% Threshold" because the short-horizon half-life is ~180 one-minute blocks. The test now keys `fee_reason` off `estimatesmartfee(60)` instead of demanding Fallback.

Qt no longer multiplies saved `nConfTarget` by 10. Those values were already 60-second block counts; *10 turned "6 minutes" into "1 hour". The dropdown snaps to {20, 40, 60, ...}.

`policyestimator_tests` has a numeric 8-10x baseRate band at one medium period again. `btq-qt.1` and the send examples say default 60.

augchan42:
## Follow-up review: the failure-counting fix goes too far, and two corrections to our earlier comment

**Corrections to our earlier comment:**
- **The Qt finding was wrong.** Labels are `n × nPowTargetSpacing` (`sendcoinsdialog.cpp:173`), so on BTQ a saved 1008 blocks meant 16.8 hours, not "1 week". Rounding up to the next `{20..10080}` entry (`:226-240`) is correct, and it gives a longer target and a lower fee, not a higher one. Removing the ×10 in 961d1fd0c was right.
- **The "under 10 blocks not counted" finding was overstated.** `blocksAgo >= scale` with scale 10 at 1-minute blocks is upstream's scale 1 at 10-minute blocks in wall-clock time.

**Medium: an unmined removal after one block now counts as a failure in every horizon** (`fees.cpp:496-504`, cf077eb57).
- **What happens:** `periodsAgo = max(1, blocksAgo / scale)` charges `failAvg[0]` for a transaction removed after 1–9 blocks in the short (scale 10), medium (20) and long (240) horizons. So a transaction that waited one minute counts as "not confirmed within 240 blocks".
- **Upstream:** it never charges a failure for a target the transaction didn't wait through. At its larger scales, removals younger than one period are dropped. Upstream's long horizon ignores removals younger than about 4 hours; this now counts them from one minute.
- **Effect:** low-fee transactions that get RBF-bumped or evicted within minutes push their bucket below the 95% success threshold in the long horizon. `estimateConservativeFee` reads the long horizon for `2*target <= 240`, so conservative estimates rise. That includes `estimatesmartfee`'s default mode for targets up to 120, including the RPC default of 60, and `-walletrbf=0` wallets. The default wallet path (economical, target 60) is not affected.
- **Rounding is also inconsistent:** 1–9 blocks round up to one period, but 19 blocks still rounds down to one.
- **Suggestion:** go back to rounding down. Retarget `sanity_check_rbf_estimates` so its RBF happens after `SHORT_SCALE` blocks. Or keep the change and document it as a deliberate divergence.

**Low: no test covers the new branch.** No unit test calls `removeTx(inBlock=false)`, and the old condition passes the current suite. Suggest a unit test at `scale - 1`, `scale` and `scale + 1`.

**Test changes look fine.** The `wallet_basic.py` change doesn't hide a regression, because economical falls back to conservative when it has no estimate (`fees.cpp:903`). The 8–10x band in `policyestimator_tests` can only fail on its lower bound, so it asserts at least 8x.

**On #221:** `ROLLING_FEE_HALFLIFE` is in seconds. `GetMinFee` decays it by `GetTime()` deltas (`txmempool.cpp:1106-1114`), so the mempool floor already decays over 12 wall-clock hours, the same as upstream. Dividing it by 10 would make it decay 10x faster than upstream. Worth checking before acting on #221.

_Checked by Codex and Opus._

PR #194, opened 2026-09-01
p2p: weight-based orphanage, fetch parents from every announcer

BarneyChambers:
## Summary
- Orphanage stored 100 txs and only asked the first announcer for missing parents.
- Multiple peers can now announce the same wtxid. A later announcer also gets parent GETDATAs.
- Eviction is by unique weight (cap `2 * MAX_BLOCK_WEIGHT`) and latency score (3000), with `MAX_STANDARD_TX_WEIGHT` reserved per peer. `-maxorphantx` still parses and is ignored.

## Test plan
- [ ] `src/test/test_btq --run_test=orphanage_tests`
- [ ] `test/functional/p2p_invalid_tx.py` (orphan handling)
- [ ] Three peers announce the same Dilithium child: parent is requested from more than the first announcer
- [ ] Confirm `-maxorphantx=1000` still starts the node and logs that the flag is ignored


BarneyChambers:
Tracking: #202

augchan42:
## Independent review — the multi-announcer feature doesn't fire on normal relay

Cross-model review (Codex, context-blind) against `.claude/code-review-lanes.md`. All findings verified against the code before posting.

### [high] Ordinary INV announcers never reach `AddAnnouncer`

`AddAnnouncer()` has exactly one non-test call site — `src/net_processing.cpp:4205` — and it sits inside the **`TX` message handler**, under:

```cpp
if (AlreadyHaveTx(GenTxid::Wtxid(wtxid))) {
    // Another peer already gave us this orphan. Record this announcer
    if (m_orphanage.HaveTx(GenTxid::Wtxid(wtxid))) {
        if (m_orphanage.AddAnnouncer(wtxid, pfrom.GetId())) {
            RequestOrphanParents(pfrom, *peer, tx);
```

In normal relay the second peer doesn't send a `TX` — it sends an **INV**. The INV path checks `AlreadyHaveTx()`, sees the orphan is already stored, and therefore never sends `getdata`, so the peer never sends the transaction and this block is never reached. `ForgetTxHash()` also clears pending requests after the orphan is first stored.

So the behaviour only triggers when a peer *unsolicitedly re-pushes the full transaction*. The scenario #202 describes — "a Dilithium child that three peers know about still depends on the first one answering" — is unchanged for ordinary announcements.

**Fix:** in the INV path, when `AlreadyHaveTx()` is true *because* the tx is a stored orphan, add the INV sender as an announcer and call `RequestOrphanParents()` with the stored transaction. Cover both `MSG_TX` and `MSG_WTX`.

### [medium] Witness-malleated variants corrupt the txid index

`src/txorphanage.cpp:57`. `m_orphans` is keyed by wtxid, and `AddTx` gates on `m_orphans.count(wtxid)` — so two transactions with the **same txid and different witnesses** are both stored. But:

```cpp
m_txid_to_wtxid.emplace(hash, wtxid);
```

`emplace` is a no-op when the key already exists, so only the first variant is reachable by txid. Erasing that one removes the sole txid mapping while the other variant stays in `m_orphans`, invisible to `ResolveWtxid`. `ProcessOrphanTx()` erases by txid, so it removes an arbitrary variant or none, and residual variants can be reconsidered repeatedly.

**Fix:** either reject a new wtxid whose txid is already indexed (restoring the old one-orphan-per-txid invariant, with `assert` on insertion), or make the index `txid -> set<wtxid>` and erase exact wtxids on the reconsideration paths.

### [medium] Announcer bookkeeping isn't charged to the limits that bound it

`OrphanTxBase::GetUsage()` returns only `GetTransactionWeight(*tx)`, so the `std::set<NodeId> announcers` is never counted in `m_total_orphan_usage` and `NeedsTrim()` never sees it. `m_total_announcements` is incremented but not bounded, and `AddAnnouncer()` does not call `LimitOrphans()`.

To be fair about the size of this: the latency cap does bound it. `GetLatencyScore()` returns at least 1 per orphan against `DEFAULT_MAX_ORPHANAGE_LATENCY_SCORE = 3000`, so orphan count is capped at ~3,000, and announcers are capped by `DEFAULT_MAX_PEER_CONNECTIONS = 125`. Worst case is ~375,000 set nodes, roughly 15 MB on top of the 16 MB payload ceiling — **bounded, but unaccounted**, and it scales with `-maxconnections` rather than with the weight budget the design is stated in terms of.

Eviction cost is worth a look too: removing a widely-shared orphan scans peers and orphans per announcement, roughly `O(P(P+3000))` for a fully shared entry.

**Fix:** charge announcements to the accounting (or cap announcers per orphan), and call `LimitOrphans()` after `AddAnnouncer()`.

### [low] Man pages still describe `-maxorphantx` as a count

Runtime help and `share/examples/btq.conf` are correctly updated to say it's deprecated and ignored, but `doc/man/btqd.1` and `doc/man/btq-qt.1` still say it keeps at most `<n>` unconnectable transactions. Same regeneration gap as #188 has.

### [low] The multi-announcer path has no regression test

`src/test/orphanage_tests.cpp:135` calls `AddAnnouncer()` **directly**, so it passes even though no INV ever reaches that code in production. A P2P-level test that sends an INV from a second peer and asserts parent requests would have caught the finding above.

### Verified correct

The weight bound scales properly: `DEFAULT_MAX_ORPHANAGE_WEIGHT = 2 * MAX_BLOCK_WEIGHT` = 16 MW versus upstream's 8 MW, correctly tracking BTQ's doubled block weight. `DEFAULT_RESERVED_ORPHAN_WEIGHT_PER_PEER = MAX_STANDARD_TX_WEIGHT` = 400,000 matches upstream. Integer arithmetic checks out — totals stay near 16.4M and latency near 3,061 before trimming, and `N x 400,000` is evaluated as `int64_t`. Changing `IteratorComparator` from pointer comparison to `a->first < b->first` is a genuine improvement in determinism. No consensus impact.

BarneyChambers:
Orphans are transactions we received whose parents are not in the mempool yet. We keep the child and ask that peer for the missing parents.

This PR already let more than one peer be recorded as an announcer. The only production call sat on the TX handler, after `AlreadyHaveTx`. Normal relay is INV (a 32-byte hash). Once the first peer stored the orphan we marked it as already-have, forgot pending GETDATAs, and later INVs never became parent requests. The Dilithium case in the description, three peers know the child and we still wait on the first, was unchanged for ordinary announcements.

INV of a stored orphan now adds that peer and sends them parent GETDATAs, MSG_TX and MSG_WTX. Same-txid different-witness inserts are refused, the txid index only kept the first variant so erase could strand the other. Announcers are capped at 10 so the set does not grow with `-maxconnections`. `LimitOrphans` runs after each new announcer.

augchan42:
## Follow-up review — main fixes verified; three issues remain

This is a big improvement and the core of it is right. Confirmed working: ordinary INV announcers now request parents, duplicate-txid witness variants are rejected, `MAX_ANNOUNCERS_PER_ORPHAN` caps announcers at ten, trimming runs after additions, and the manpages are updated. `NeedsTrim` implements #202's `min(2 * MAX_BLOCK_WEIGHT, n_peers * MAX_STANDARD_TX_WEIGHT)` and the `1 + inputs/10` latency score as specified.

**1. The overflow test no longer reaches a limit.** `test/functional/p2p_invalid_tx.py:154` sends 101 fixtures and expects `orphanage overflow, removed 1 announcements`. Under the new limits it never fires. `SCRIPT_PUB_KEY_OP_TRUE` is 31 bytes, so each fixture serializes to 91 bytes — 1,456 WU at WSF 16, and 147,056 WU for all 101. That is well under one peer's 400,000 WU reserve, and the latency score of 101 is well under the 3,000 cap.

The test workload has to grow. Adding a second announcing peer does not help and makes it worse: `AddAnnouncer` increments only that peer's `m_total_usage` and `m_total_announcements`, never `m_total_orphan_usage` or `m_total_latency_score`, while `n_peers` going to 2 raises `reserved` from 400,000 to 800,000.

Repeating this 101-fixture batch with distinct orphans from a second peer still does not help: it adds 147,056 WU but raises the weight allowance by 400,000 WU. Distinct orphans do increase the global latency score, but two such batches reach only 202 against the 3,000 cap.

With one contributing peer you need to cross one of two thresholds:

- **Weight:** more than 400,000 WU of unique orphans. At the current 1,456 WU per fixture that is 275 of them. Fewer, heavier fixtures are tidier — `AddTx` accepts an orphan up to `MAX_STANDARD_TX_WEIGHT`, so a handful of large ones gets there without a 275-element loop.
- **Latency:** a total score above 3,000, where each orphan scores `1 + inputs/10`.

**2. Trimming a shared announcement frees no global budget.** In `LimitOrphans`, the shared branch erases the announcer and decrements `m_total_announcements` and that peer's `m_total_usage`, but leaves `m_total_orphan_usage` and `m_total_latency_score` untouched. `NeedsTrim` reads those two. So when the trim was triggered by latency, each pass does a full `m_orphans` scan to find a candidate and changes nothing the loop condition looks at.

It does terminate — a peer's usage eventually reaches zero and the random fallback takes over — but it can scan the whole map many times to get there. Worse, dropping a peer's usage to zero lowers `n_peers`, which lowers `reserved`, which makes `NeedsTrim` *more* likely to stay true. Consider skipping the shared branch when the trim was triggered by latency, or tracking a candidate set instead of rescanning.

**3. Pending work is lost when the chosen announcer disconnects.** `AddChildrenToWorkSet` assigns a child to one random announcer. `EraseForPeer` then erases `m_peer_orphanage_info[peer]` at the top — work set included — while the orphan itself survives if other announcers remain. Nobody reassigns the work, so the child sits until it expires even though a connected peer announced it. Reassigning to a remaining announcer on disconnect should be enough.

None of these block the design. (1) is the one that will show up in CI.


---

augchan42:
## Follow-up review — main fixes verified; three issues remain

Normal INV announcers now reach parent requests, same-txid witness variants are rejected, announcers are capped at ten, trimming runs after additions, and the manpages are updated.

- **Medium:** `p2p_invalid_tx.py:154` still sends 101 fixtures and expects overflow. They total about 147,056 WU, below one peer’s 400,000-WU reservation and the 3,000 latency cap, so the test no longer reaches either limit.
- **Medium:** trimming can repeatedly scan the full orphan map while removing shared announcements that free no global weight or latency budget.
- **Medium:** child reconsideration is assigned to one random announcer. If that peer disconnects, `EraseForPeer` deletes its work set but retains the shared orphan without reassigning pending work to another announcer.

Please retarget the functional overflow fixture and add coverage for INV-only additional announcers, shared-announcer trimming, and disconnect/reassignment.

augchan42:
**Review findings**

**1. Medium: an attacker can force eviction of honest peers' orphans**

In `LimitOrphans` (`src/txorphanage.cpp:214-256`), the victim peer is chosen purely by weight (`m_total_usage`). If that peer has no uniquely-announced orphan and the latency score is over the cap, the code erases a *shared* orphan outright, removing it for every other announcer too.

The new inv path (`src/net_processing.cpp:3899-3905`) makes a peer an announcer, charged the orphan's full weight, just for sending an `inv`. An attacker can:
1. From connection A, `inv` the hashes of honest peers' large Dilithium orphans (about 60k weight each at `WITNESS_SCALE_FACTOR` 16). A becomes the heaviest peer, and everything it holds is shared.
2. From connections B through I, push the latency score over 3000 with minimum-size orphans (about 1k weight each).
3. From then on, every trim picks A and erases an honest orphan entirely. The attacker's tiny orphans survive, and one more keeps latency above the cap.

A simpler variant: spread tiny orphans across enough connections that an honest peer with one large orphan becomes the heaviest peer. Upstream scores peers by `max(weight share, latency share)` and only evicts announcements, which avoids this.

**2. Low: inv-only peers can fill all announcer slots**

`MAX_ANNOUNCERS_PER_ORPHAN = 10` slots are first-come-first-served, including for peers that only sent an `inv`, and `RequestOrphanParents` runs only when `AddAnnouncer` succeeds. Nine attacker connections can fill the slots on every new orphan, so honest announcers are never asked for parents. That defeats the "fetch parents from every announcer" goal.

**3. Test: `announcer_and_weight_trim` passes without checking anything**

Each `MakeOrphanTx` is about 1376 weight, which is over `reserved_usage_per_peer=400`, and with one peer the effective limit is 400. So `LimitOrphans` evicts everything, and `TotalOrphanUsage() <= 2000` holds trivially at 0. The test doesn't exercise the per-peer limit or check that a shared orphan survives a trim.

**Unverified**
- `LimitOrphans` runs once per matching inv entry under `cs_main`, so one large inv could trigger thousands of full scans.
- `feature_rbf.py`, `mempool_packages.py` and `mempool_package_onemore.py` still pass `-maxorphantx`, which is now ignored.

Accounting, termination, iterator safety, locking and txrequest dedup all look correct.

_Checked by Opus._

BarneyChambers:
LimitOrphans no longer deletes a shared orphan.

Trim used to EraseTx the victim peer's tx even when other peers still announced it, so an inv-only flood could drop honest Dilithium orphans. Unique orphans are evicted first. Shared ones only lose that announcer. Parent fetch is no longer gated on a successful AddAnnouncer.

If peer usage counters are already 0, subtracts saturate instead of wrapping to UINT_MAX, so the loop stays in the empty-accounting fallback until usage is under the cap. Tests cover the inv-flood case, tx-provider displacement of an inv-only slot, and the zeroed-counter fallback.

BarneyChambers:
Pushed a follow-up for leftover DoS shape in the orphanage:

- `LimitOrphans` randomizes among max-metric peers and among that peer's unique/shared orphans, instead of always taking the lowest NodeId / wtxid.
- INV parent fetch and trim run only after `AddAnnouncer` succeeds, so same-peer and cap-rejected re-INV cannot walk every parent per INV item.
- Unit test: two equal-usage peers, 32 seeds, both uniques survive across seeds.

augchan42:
## Follow-up review: shared-orphan erase fixed; new medium in the latency path

**Fixed:**
- **Shared orphans:** `LimitOrphans` no longer erases one; it strips an announcer instead (`txorphanage.cpp:290-335`). `inv_announcer_does_not_erase_shared_orphan` would fail on the old code.
- **Unit test:** `announcer_and_weight_trim` now checks usage before and after trimming.
- **Tier 2 items:** the INV path reaching `AddAnnouncer`, same-txid witness variants, the man page, and work reassignment on disconnect are all fixed.
- **Termination:** each pass removes an orphan or an announcement, and shared orphans become unique and then get erased, so usage always comes back under the limit.
- **Locking:** `cs_main` is held where required, and `m_rng` is used under `g_msgproc_mutex`.

**Medium: under latency pressure, honest orphans are erased as soon as they arrive** (`txorphanage.cpp:255-270`). When `m_total_latency_score > m_max_latency_score`, the code erases the smallest *unique* orphan in the whole orphanage, whoever announced it. Shared orphans are skipped while any unique orphan exists.
- **Attack, with about 10 attacker connections:**
  1. Send about 3,000 minimum-size standard orphans (about 1,312 WU each) from one connection, and INV each from a second, so every attacker orphan is shared.
  2. Latency reaches 3,000. Total weight is about 3.94 MWU, which is under the 10 × 400k limit.
  3. An honest peer sends orphan X. `LimitOrphans` runs immediately (`net_processing.cpp:4292`), X is the only unique orphan, and it is erased before a second announcer can arrive.
  4. The attacker refills after the 20-minute expiry, at about 250 KB each time.
- **Fix:** under latency pressure, pick the peer with the highest latency score first, then evict among that peer's orphans, the same way weight mode does.
- The new test `inv_announcer_does_not_erase_shared_orphan` encodes the unique-first order, so it would need updating.

**Low: INV announcers refused at the 10-slot cap.** `txorphanage.cpp:96` refuses them. Push-out only happens for peers that send the full tx, and honest peers relay by INV. f22c5cf42 made the INV parent fetch depend on being added again (`net_processing.cpp:3903-3906`), so honest INV relayers blocked by the cap are never asked for parents.

**Low:**
- **Fuzz:** `fuzz/txorphan.cpp` never asserts usage or latency are back under the limits after `LimitOrphans`. Negative fuzz peer ids reach the `worst < 0` fallback, which production can't reach. The fixed RNG seed means the random tie-breaks aren't fuzzed.
- **Trim cost:** each trim pass rescans all of `m_orphans` under `cs_main`, so cost is O(evictions × orphans) when the limit drops by several peers at once.
- **`SaturatingSub`** hides accounting desync in release builds; an `Assume(value >= delta)` would catch it.

**Test gap:** no functional test sends an INV from a second peer and checks that parents are requested.

_Checked by Codex and Opus._

PR #212, opened 2026-09-14
test: fix 43 functional test failures from the block subsidy, witness scale and genesis hash (BTQ-115)

augchan42:
Part of BTQ-115 (#104). This PR fixes 43 test variants and changes only test code.

## Result

Full suite, run with the same arguments as the functional-tests job: 304 tests, **246 passed, 49 failed (all 49 known, 0 new), 9 skipped**. `known_failures.txt` shrinks from 92 tests on master to 49.

`check_known_failures.py` exits 0 with no "listed test passed" warnings. Each commit removes the `known_failures.txt` lines for the tests it fixes.

## Commits

1. **feature_p2mr.py.** The test expected an older wording of the node's "Dilithium opcodes are only available in P2MR tapscript" error (`script_error.cpp:115`).
2. **wallet_txn_clone / wallet_txn_doublespend.** Derive amounts from `block_subsidy()`, the pattern from #141.
3. **Genesis-hash literals.** Three tests compare against Bitcoin's regtest genesis hash (`0f9188f1…2206`) and now ask the node with `getblockhash(0)`.
4. **Subsidy changes in wallet tests** (18 files).
5. **Subsidy changes in feature, mempool, p2p and rpc tests** (7 files).
6. **Other consensus constants:**
   - `wallet_groups.py` hard-codes fees for Bitcoin's witness scale factor of 4. BTQ's is 16.
   - `mempool_accept.py` builds its oversized scriptSig against Bitcoin's 1650-byte `MAX_STANDARD_SCRIPTSIG_SIZE`. BTQ's limit is 15,000.

Amounts follow #141: tests use `block_subsidy()` where they check a block reward, and scale other amounts and hand-set fees to a tenth. Where a tenth doesn't work, a comment in the test says why. Most of those cases come from `WITNESS_SCALE_FACTOR` 16: transactions weigh fewer vbytes, so fee rates move, and many-output transactions go over `MAX_STANDARD_TX_WEIGHT`. The commit messages list each one.

## Not in this PR

**`wallet_send.py`, `rpc_psbt.py` and `wallet_fundrawtransaction.py`** get past the subsidy failures but then hit a node bug. `src/wallet/rpc/spend.cpp:690` has `CHECK_NONFATAL(min_input_weight == 165)`, which is Bitcoin's 41 bytes × 4 + 1. At BTQ's witness scale factor the minimum is 41 × 16 + 1 = 657. As a result, any `fundrawtransaction`, `walletcreatefundedpsbt` or `send` call that passes an input `weight` returns "Internal bug detected". These three tests' changes will go in a separate PR together with the node fix, so they can be tested end to end.

Other known failures that still need work:

- **`p2p_headers_sync_with_minchainwork.py`.** After the genesis fix it gets further, then fails with `time-too-new`: mining 4110 blocks runs past BTQ's 15-minute `MAX_FUTURE_BLOCK_TIME` (Bitcoin's is 2 hours).
- **`wallet_fundrawtransaction.py` `test_22670`.** Even after the node fix it needs a new scenario: at witness scale factor 16, two P2WPKH inputs don't cost enough in fees for the test's bounds.
- **`rpc_psbt.py` and BTQ-110.** The `not(19 == 2)` failure was the subsidy, not P2MR. The bech32m cases in `rpc_psbt` use taproot, not P2MR, and they pass.

## Testing

- Ran every changed test directly, in each wallet variant.
- Ran the full suite: `test_runner.py --ci --quiet --jobs 8 --timeout-factor=4`, then `check_known_failures.py`.
- Ruff finds the same issues in the changed files on master and on this branch, so none are new.


augchan42:
## Reviewer's guide

30 files, six commits. Each commit covers one category and removes a block of known failures. Reviewing per commit is smaller than reviewing the combined diff.

| Commit | Removes | Files | What to check |
|---|---|---|---|
| `a1f1ad3b8` P2MR error message | 1 | 2 | The node's wording at `src/script/script_error.cpp:115`. The test was changed to match the node, so it is correct only if the node's message is correct. |
| `f16a0b6ce` txn_clone / doublespend amounts | 6 | 3 | Amounts come from `block_subsidy()` instead of a literal. |
| `13bccb6e1` genesis hash | 1 | 4 | `getblockhash(0)` instead of Bitcoin's `0f9188f1…2206`. |
| `50485e2b7` wallet amounts | 27 | 17 | The largest commit. The same pattern repeats across all 17 files. |
| `8568bf853` feature, mempool, p2p, rpc amounts | 6 | 8 | Same pattern as above. |
| `7479ab29d` witness scale and scriptSig limit | 3 | 3 | Fee and weight arithmetic. The 15,000 is `MAX_STANDARD_SCRIPTSIG_SIZE` (`src/policy/policy.h:52`), which is `MAX_SCRIPT_ELEMENT_SIZE`; line 55 asserts it against `WITNESS_SCALE_FACTOR`. |

Checks already run:

- **The count.** `known_failures.txt` loses 44 lines and gains 1: `p2p_headers_sync_with_minchainwork.py` now fails with `time-too-new` instead of the genesis mismatch, so it stays listed. Net 43.
- **No test weakened.** 88 assertion lines removed, 89 added, no skips added.
- **Values derived, not pasted.** 18 new uses of `block_subsidy()` and 3 of `WITNESS_SCALE_FACTOR`. Where a literal remains, the commit message says why.
- **No new failures.** The functional job passes, and it errors on any failing test that is not listed.

To check the count directly:

```
git diff master...7479ab29d -- test/functional/known_failures.txt | grep -c '^-[^-]'   # 44
git diff master...7479ab29d -- test/functional/known_failures.txt | grep -c '^+[^+]'   # 1
```

Reasons to reject a hunk: an exact assertion turned into a range or `>=`, a skip added, or a bare `5` where `block_subsidy()` belongs, which breaks at the first halving.

No `src/` files change here, so there is no node risk in this PR. That starts at #213.

test/functional/feature_p2mr.py
```diff
@@ -297,7 +297,7 @@ def test_05_dilithium_error_messages(self):
         print(sub('5a. P2TR must return exact error message'))
         print(why('The error message distinguishes "Dilithium blocked in tapscript" from other failures.'))
         print(how('Attempt OP_CHECKSIGDILITHIUM in P2TR and check the exact error string.'))
-        expected_msg = 'Dilithium opcodes are not available in tapscript'
+        expected_msg = 'Dilithium opcodes are only available in P2MR tapscript (witness v2)'
```
augchan42, line 300: This changes the test to match the node. The node's wording is at `src/script/script_error.cpp:115`. The test is therefore only correct if that message is correct, which is a call for you rather than something I can settle in a test change. If the message should read differently, the node and this line change together.

test/functional/mempool_accept.py
```diff
@@ -296,7 +303,9 @@ def run_test(self):
             rawtxs=[tx.serialize().hex()],
         )
         tx = tx_from_hex(raw_tx_reference)
-        tx.vin[0].scriptSig = CScript([b'a' * 1648]) # Some too large scriptSig (>1650 bytes)
+        # BTQ's MAX_STANDARD_SCRIPTSIG_SIZE is MAX_SCRIPT_ELEMENT_SIZE, so one
+        # full-size push plus its PUSHDATA2 prefix is too large.
+        tx.vin[0].scriptSig = CScript([b'a' * MAX_SCRIPT_ELEMENT_SIZE])
```
augchan42, line 308: Upstream hard-codes 1648 against Bitcoin's 1650-byte limit. BTQ's `MAX_STANDARD_SCRIPTSIG_SIZE` is `MAX_SCRIPT_ELEMENT_SIZE`, 15,000 (`src/policy/policy.h:52`), and line 55 asserts that value against `WITNESS_SCALE_FACTOR`. A full-size push carries a 3-byte PUSHDATA2 prefix, so the scriptSig is 15,003 bytes and the reject reason is still `scriptsig-size`. Using the constant rather than a literal keeps this correct if the limit moves.

test/functional/wallet_groups.py
```diff
@@ -54,109 +62,111 @@ def run_test(self):
...
         assert_equal(input_addrs[0], input_addrs[1])
         # Node 2 enforces avoidpartialspends so needs no checking here
 
-        tx4_ungrouped_fee = 2820
-        tx4_grouped_fee = 4160
-        tx5_6_ungrouped_fee = 5520
-        tx5_6_grouped_fee = 8240
+        # Lower than upstream's because BTQ's witness scale factor is 16, not
+        # 4, so the P2WPKH inputs these transactions spend weigh fewer vbytes.
+        tx4_ungrouped_fee = 2400
```
augchan42, line 134: These are lower than upstream's because a P2WPKH input weighs fewer vbytes at witness scale 16. The paired `-maxapsfee` values at line 37 are one satoshi below and exactly at the tx5/tx6 difference, 6220 - 4320 = 1900, so nodes 3 and 4 take opposite branches. If you change either fee, both `-maxapsfee` values have to move with it.

test/functional/wallet_groups.py
```diff
@@ -168,11 +178,15 @@ def run_test(self):
         self.generate(self.nodes[0], 1)
 
         self.log.info("Fill a wallet with 10,000 outputs corresponding to the same scriptPubKey")
-        for _ in range(5):
-            raw_tx = self.nodes[0].createrawtransaction([{"txid":"0"*64, "vout":0}], [{addr2[0]: 0.05}])
+        # Upstream sends 5 transactions of 2000 outputs. At BTQ's witness
+        # scale factor of 16 that many outputs weigh more than
+        # MAX_STANDARD_TX_WEIGHT, so send 20 of 500, each the weight upstream's
+        # transactions had.
+        for _ in range(20):
```
augchan42, line 185: Upstream sends 5 transactions of 2,000 outputs. At witness scale 16 that exceeds `MAX_STANDARD_TX_WEIGHT`, so this sends 20 of 500. The total is still 10,000 outputs, and each transaction carries the weight upstream's did.

augchan42 (commented):
Notes on the four hunks that are not a repeat of the same pattern, so they do not have to be worked out from the diff.

augchan42:

Two independent reviews (Codex blind, and Opus) found no problems.
- **Scope:** only files under `test/functional/` change.
- **Assertions:** none were deleted or loosened, and nothing new is skipped.
- **Numbers:** re-derived subsidy (5 BTQ, halving at 1500), vsize/fee (ceil(w/16), e.g. the `wallet_groups` 2400/3360/4320/6220 sat values) and many-output weight figures all match BTQ's constants.
- **CI:** all 43 removed `known_failures.txt` entries show Passed.

**Nit:** a few checks now read the genesis hash from the node (`getblockhash(0)`) instead of hard-coding it. The startup assert in `chainparams.cpp` covers this, but a literal would be stricter.

_Checked by Codex and Opus._

PR #213, opened 2026-09-14
wallet: require BTQ's minimum input weight of 657 in input_weights (BTQ-115)

augchan42:
Stacked on #212. When #212 merges, GitHub retargets this PR to master. Part of BTQ-115 (#104).

## The bug

`fundrawtransaction`, `walletcreatefundedpsbt` and `send` check each `input_weights` entry against the weight of an empty input (`src/wallet/rpc/spend.cpp:690`):

```cpp
const int64_t min_input_weight = GetTransactionInputWeight(CTxIn());
CHECK_NONFATAL(min_input_weight == 165);
```

165 is Bitcoin's 41 bytes × 4 + 1. BTQ's `WITNESS_SCALE_FACTOR` is 16, so an empty input weighs 41 × 16 + 1 = **657**. The `CHECK_NONFATAL` can never pass, so any of these calls that passes an input weight returns `Internal bug detected: min_input_weight == 165`.

The fix keeps upstream's check and its error message and changes the numbers: 657, and "* 16 (witness scaling factor)". That keeps the diff against upstream as small as possible.

## Tests

This fixes `wallet_send.py`, `rpc_psbt.py` and `wallet_fundrawtransaction.py`, in both wallet variants, and removes their six lines from `known_failures.txt`.

The test changes follow #141 and #212: amounts come from the block subsidy, and fees and vsizes from the witness scale. Two fixes in `wallet_fundrawtransaction.py` are for intermittent failures, so I measured them rather than rely on reruns:

- **The fee comparison tests failed in 1 of my 14 reruns.** At upstream's amounts, the outputs total 4.2 of a 5 BTQ block reward. Sampled 400 times at the same point in the test, coin selection added a second input in 16. When `fundrawtransaction` and `sendmany` picked differently, their fees were 48 sats apart, one P2WPKH input, against a 12-sat tolerance. At a tenth of the amounts, all 400 used a single input.
- **`test_22670` needed P2PKH inputs, and then failed about 2% of the time.**
  - At witness scale 16, two P2WPKH inputs don't cost enough in fees to leave a target between the test's bounds, so the tester wallet is now funded with P2PKH coins.
  - The wallet pays exactly 0.1 BTQ/kvB, which is also `testmempoolaccept`'s default maximum fee rate.
  - Funding, signing and checking each bound 150 times gave `max-fee-exceeded` 10 times out of 600. Every one was a transaction whose signature came out a byte shorter than the wallet's estimate.
  - The check now passes `maxfeerate=0`, keeping the minimum-relay-fee check the test is about. All 600 passed with it.
  - Upstream's version of this test is unchanged; P2WPKH at scale 4 rarely loses a vbyte this way.

After both fixes, `wallet_fundrawtransaction.py` passed 16 of 16 runs, 8 in each wallet variant.

## Testing

- Rebuilt `btqd` and ran each changed test in both variants.
- Ran the full suite with the functional-tests job's arguments, then `check_known_failures.py`: 304 tests, **252 passed, 43 failed (all 43 known, 0 new), 9 skipped**. With #212, `known_failures.txt` goes from 49 tests to 43.
- Ruff finds nothing new in the changed files, and the whitespace lint is clean.


augchan42:

Two independent reviews (Codex blind, and Opus) found no problems.
- **657 is correct:** it is the smallest input weight under WSF 16 (41 × 15 + 41 + 1).
- **Coverage:** every RPC that accepts `input_weights` goes through `FundTransaction`.
- **Test values:** the P2WPKH 48 vB and the batched-send weights re-derive correctly.

**Nit, predating this PR:** the `weight` help text at `spend.cpp:793` still mentions only 73-byte ECDSA signatures and doesn't say the scale is 16.

_Checked by Codex and Opus._

PR #215, opened 2026-09-15
wallet: only treat a descriptor's own scripts as mine

augchan42:
Stacked on #214. Part of BTQ-115 (step 2 in the known-failures plan).

## Problem

`feature_segwit.py` failed about half the time with:

```
Internal bug detected: Fee needed > fee paid
```

This is a wallet bug, not a test bug. Any descriptor wallet that holds one key in more than one descriptor can hit it when sending.

## Cause

`DescriptorScriptPubKeyMan::IsMine` had a BTQ-only fallback. It returned `ISMINE_SPENDABLE` for any P2WPKH script whose key the descriptor holds, even when the descriptor never produces that script. Upstream `IsMine` only checks `m_map_script_pub_keys`.

`MarkUnusedAddresses` adds every script that `IsMine` accepts to `m_map_script_pub_keys`. feature_segwit imports one key K as `sh(wsh(multi(1,K)))`, `wsh(multi(1,K))`, `sh(wpkh(K))` and `wpkh(K)`. After a payment to `wpkh(K)`, the three other descriptors track that script too.

When coin selection sizes an input, `CWallet::GetSolvingProvider` returns the provider of the first descriptor that claims the script. Descriptors are ordered by ID, which differs per wallet. The multisig descriptors' providers have no public key for K, so `InferDescriptor` returns `addr()`. The input then can't be sized, and coin selection counts no fee for it. The final size estimate merges every descriptor's provider and does count the input, so `CreateTransaction` stops at the fee check.

The fallback came in with 43542f662 (Dilithium wallet, phase 3). It stayed for ECDSA keys when 29a82ac1a removed the Dilithium part.

## Fix

Remove the fallback, so `IsMine` matches upstream. `m_map_script_pub_keys` is rebuilt from the descriptors when a wallet loads, so existing wallets need no migration.

## Tests

- New `ismine_tests` case: a `wsh(multi(1,K))` descriptor does not claim P2WPKH(K), and `MarkUnusedAddresses` does not add it. Fails before the fix, passes after.
- Wallet and Dilithium unit suites pass (160 cases).
- A scratch script imports one key into the four descriptors above, pays the `wpkh` address, and checks `listunspent`'s `solvable` field. Before the fix, 11 of 24 wallets could not size their own coin. After, 0 of 24.
- `feature_segwit.py`: each of the three variants passed 5 of 5 runs. A temporary log in `CreateTransactionInternal` showed no unsized inputs. The log is not part of this PR.
- Full functional suite with `check_known_failures.py`: 304 tests, 256 passed, 39 failed (all 39 on the list, 0 new), 9 skipped.

## known_failures.txt

Removes the three `feature_segwit.py` entries. The legacy-wallet variant was listed with the other two. It never showed the fee error, since legacy wallets don't use `DescriptorScriptPubKeyMan`. Its earlier CI failures were the `errors` key check in `send_to_witness`, which #214 fixed.


augchan42:
## Review: no findings, ready to merge (release note requested)

Two independent reviews (Codex blind, and Opus) found no bugs. `DescriptorScriptPubKeyMan::IsMine` now matches upstream. P2MR (`GetTrackedP2MRScriptIsMine`), Dilithium, legacy-migrated and ranged descriptors are unaffected.

**Release note needed:** the PR says existing wallets need no migration. That's true of the script map but not of the balance:
- **Who's affected:** a descriptor wallet that holds key K only in a descriptor that doesn't produce P2WPKH (imported `pkh(K)`, `sh(wpkh(K))`, `tr(K)`, a multisig) and has received coins at P2WPKH(K).
- **What changes:** those coins drop out of `getbalance` and `listunspent`.
- **Recovery:** they aren't lost; `importdescriptors` of `wpkh(K)` plus a rescan brings them back.

The fallback shipped from v0.1.0-testnet, so such wallets may exist. A one-line note would cover it.

_Checked by Codex and Opus._

